PR #2, opened 2018-05-14
Utils for installing features from Maven repo

ericglau:
Add utils for use with https://github.com/WASdev/ci.maven/issues/330


src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,559 @@
...
+            debug("Features were listed for the plugin. Skipping server.xml.");
+        }
+        return result;
+    }
+    
+    private Set<String> getConfigDropinFeatures(File serverDirectory, String folderName){
+        Set<String> result = new HashSet<String>();
+        String configDropin = serverDirectory.getAbsolutePath() + "/configDropins/";
+        File configDropinFolder = new File(configDropin+folderName);
+        String[] overrideFileList = configDropinFolder.list();
```
chyt: You will have to list the files in alphabetical order. There is no guarantee on the file order from the `list()` method. According to [this doc](https://www.ibm.com/support/knowledgecenter/en/SSEQTP_8.5.5/com.ibm.websphere.wlp.doc/ae/twlp_setup_dropins.html) the files will be parsed in alphabetical order which will have an impact on which features are overwritten.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,559 @@
...
+            String onConflict = node.getAttribute("onConflict");
+            boolean onConflictMerge = Pattern.compile(Pattern.quote("MERGE"), Pattern.CASE_INSENSITIVE).matcher(onConflict).find();
+            boolean onConflictReplace = Pattern.compile(Pattern.quote("REPLACE"), Pattern.CASE_INSENSITIVE).matcher(onConflict).find();
+            boolean onConflictIgnore = Pattern.compile(Pattern.quote("IGNORE"), Pattern.CASE_INSENSITIVE).matcher(onConflict).find();
+            if (!(onConflictMerge || onConflictReplace || onConflictIgnore)){
+                onConflictMerge = true;
+            }
+            if (!onConflictIgnore){
+                if (onConflictMerge){
+                    result.addAll(getServerXmlFeatures(fileContext, includeFileName, updatedParsedXmls));
```
chyt: I'm not familiar with how feature resolution in Liberty works, but what happens if there are multiple versions of the same feature that are merged? Will you need to handle that?

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,559 @@
...
+        String includeFileName = node.getAttribute("location");
+        if (!updatedParsedXmls.contains(includeFileName)){
+            String onConflict = node.getAttribute("onConflict");
+            boolean onConflictMerge = Pattern.compile(Pattern.quote("MERGE"), Pattern.CASE_INSENSITIVE).matcher(onConflict).find();
+            boolean onConflictReplace = Pattern.compile(Pattern.quote("REPLACE"), Pattern.CASE_INSENSITIVE).matcher(onConflict).find();
+            boolean onConflictIgnore = Pattern.compile(Pattern.quote("IGNORE"), Pattern.CASE_INSENSITIVE).matcher(onConflict).find();
+            if (!(onConflictMerge || onConflictReplace || onConflictIgnore)){
+                onConflictMerge = true;
+            }
+            if (!onConflictIgnore){
```
chyt: Where do you check in the code that a conflict actually occurs? Shouldn't you still add the features if there is no conflict?

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,559 @@
...
+                if (onConflictMerge){
+                    result.addAll(getServerXmlFeatures(fileContext, includeFileName, updatedParsedXmls));
+                } else if (onConflictReplace){
+                    result = getServerXmlFeatures(fileContext, includeFileName, updatedParsedXmls);
```
chyt: If you are completely replacing the feature list, then the order in which the includes are parsed is very important. I think you're parsing the nodes in the order in which they appear in the document, so the last item will always have priority. It also means that one `replace` can completely overwrite the result of many `merge`s that previously took place.

Please verify that the order in which you are checking the `server.xml`, `overrides` and `defaults` folder, and `include` items is identical to how they are resolved Open Liberty and WLP runtimes.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,559 @@
...
+                    continue;
+                }
+                String latestJarVersion = extractVersion(latestJar.getName(), installMapPrefix, installMapSuffix);
+                if (latestJarVersion == null) {
+                    // jar without version is the oldest jar 
+                    latestJar = jar;
+                    continue;
+                }
+                String jarVersion = extractVersion(jar.getName(), installMapPrefix, installMapSuffix);
+                if (jarVersion != null && jarVersion.compareTo(latestJarVersion) > 0) {
```
chyt: I don't think that `compareTo` will work either on version comparisons (I'm not sure how the versioning works for these jars though). For example, `"1.10".compareTo("1.9") > 0` will return `false` even though 1.10 is a later version than 1.9. See the previous comment.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,559 @@
...
+        File[] installMapJars = dir.listFiles(new FilenameFilter() {
+            @Override
+            public boolean accept(File dir, String name) {
+                return name.startsWith(installMapPrefix) && name.endsWith(installMapSuffix);
+            }
+        });
+
+        File latestJar = null;
+        if (installMapJars != null && installMapJars.length > 0) {
+            for (File jar : installMapJars) {
```
chyt: The logic in this loop is very confusing. There are a bunch of solutions online on how to do this. See [here](https://stackoverflow.com/questions/198431/how-do-you-compare-two-version-strings-in-java) for an example.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,559 @@
...
+     * @param serverDirectory The server directory containing the server.xml
+     * @param noPluginListedFeatures true if there were no features listed in the plugin configuration
+     * @return the set of features that should be installed from server.xml
+     */
+    public Set<String> getServerFeatures(File serverDirectory, boolean noPluginListedFeatures) {
+        // parse server.xml features only if there are no configured features in the pom
+        Set<String> result = new HashSet<String>();
+        if (noPluginListedFeatures) {
+            debug("No features were listed for the plugin. Using server.xml.");
+            result.addAll(getConfigDropinFeatures(serverDirectory, "overrides"));
```
chyt: Shouldn't this order be reversed (parse `overrides` after `defaults`) since the item that is parsed last will overwrite previous items?

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,560 @@
...
+
+            Collection<?> resolvedFeatures = (Collection<?>) mapBasedInstallKernel.get("action.result");
+            if (resolvedFeatures == null) {
+                debug("action.exception.stacktrace: "+mapBasedInstallKernel.get("action.exception.stacktrace"));
+                String exceptionMessage = (String) mapBasedInstallKernel.get("action.error.message");
+                throw new PluginExecutionException(exceptionMessage);
+            } else if (resolvedFeatures.isEmpty()) {
+                debug("action.exception.stacktrace: "+mapBasedInstallKernel.get("action.exception.stacktrace"));
+                String exceptionMessage = (String) mapBasedInstallKernel.get("action.error.message");
+                if (exceptionMessage.contains("CWWKF1250I")){
```
chyt: What does this error code mean? Why do we only warn if we see this error message?

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,560 @@
...
+     * @param serverDirectory The server directory containing the server.xml
+     * @param noPluginListedFeatures true if there were no features listed in the plugin configuration
+     * @return the set of features that should be installed from server.xml
+     */
+    public Set<String> getServerFeatures(File serverDirectory, boolean noPluginListedFeatures) {
+        // parse server.xml features only if there are no configured features in the pom
+        Set<String> result = new HashSet<String>();
+        if (noPluginListedFeatures) {
+            debug("No features were listed for the plugin. Using server.xml.");
+            result.addAll(getConfigDropinFeatures(serverDirectory, "overrides"));
```
chyt: Also, does adding each result set from the `overrides`, `server.xml`, and `defaults` make sense? For example, if there is a conflict between `overrides` and `server.xml`, don't you want the `overrides` features to replace the `server.xml` features so you don't end up installing extra features?

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,560 @@
...
+            @Override
+            public Map<String, Object> run() throws Exception {
+                ClassLoader loader = new URLClassLoader(new URL[] { installJarFile.toURI().toURL() }, null);
+                Class<Map<String, Object>> clazz;
+                clazz = (Class<Map<String, Object>>) loader.loadClass("com.ibm.ws.install.map.InstallMap");
+                return clazz.newInstance();
+            }
+        });
+        if (mapBasedInstallKernel == null){
+            debug("mbik is null");
```
chyt: Can we have a more useful debug message here?

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,560 @@
...
+                }
+            }
+        } catch (PrivilegedActionException e) {
+            throw new PluginExecutionException("Could not load the jar " + getMapBasedInstallKernelJar(installDirectory).getAbsolutePath(), e);
+        }
+    }
+    
+    private Map<String, Object> createMapBasedInstallKernelInstance(File installDirectory) throws PrivilegedActionException {
+        final File installJarFile = getMapBasedInstallKernelJar(installDirectory);
+        String installJarFileSubpath = installJarFile.getParentFile().getName() + File.separator + installJarFile.getName();
```
chyt: Add a null check on `installJarFile` since `getMapBasedInstallKernelJar` can return null.

chyt: Actually I realize you do the null check in the constructor. In that case can you set `installJarFile` as a class variable so that you don't have to call it throughout?

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: For configDropins, we verified (using installUtility) that features do not get overridden, but are added to the set.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: We pass the entire list to the resolver, and it is up to the resolver to determine.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: If ignore is specified, we verified that the entire file is ignored.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: We will provide unit tests to verify the ordering.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: This indicates that the features are already installed, so no action is needed.  Changed the message to info.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: Since features are merged even from overrides, we can add them all into the set.

ericglau:
@chyt I've updated the code to address all existing comments, except we will still be adding unit tests for parsing server.xml (including includes and configDropins).  Also added a message on completion, running productInfo validate, and supporting the "to" parameter.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
chyt: @ericglau Does this happen even if there is no conflict? I would not expect that setting `onConflict="ignore"` is equivalent to not having the `include` at all.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
chyt: @ericglau Where is this test? I don't see it yet.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,683 @@
...
+        Set<String> result = new HashSet<String>();
+        result.addAll(getConfigDropinFeatures(serverDirectory, "overrides"));
+        result.addAll(getServerXmlFeatures(new File(serverDirectory, "server.xml"), null));
+        result.addAll(getConfigDropinFeatures(serverDirectory, "defaults"));
+        return result;
+    }
+    
+    private static Set<String> getConfigDropinFeatures(File serverDirectory, String folderName) {
+        Set<String> result = new HashSet<String>();
+        String configDropins = serverDirectory.getAbsolutePath() + "/configDropins/" + folderName;
```
chyt: Replace `getAbsolutePath()` with `getCanonicalPath()` throughout. For example, `/a/b/c/../` is an absolute path while `/a/b/` is the equivalent canonical path. If you are doing file comparisons you will want to compare the canonical files.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,683 @@
...
+        return result;
+    }
+    
+    private static Set<String> parseIncludeNode(Set<String> origResult, File serverFile, Element node,
+            List<File> updatedParsedXmls) {
+        Set<String> result = origResult;
+        String includeFileName = node.getAttribute("location");
+        File includeFile = new File(includeFileName);
+        if (!includeFile.isAbsolute()){
+            includeFile = new File(serverFile.getParentFile().getAbsolutePath(), includeFileName);
```
chyt: Here, once again get the canonical file of this file. So probably replace with 

```
new File(serverFile.getParentFile().getAbsolutePath(), includeFileName).getCanonicalFile()
```

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,683 @@
...
+        }
```
chyt: Else, `includeFile` is absolute, so you will also want the canonical file. So add an else block setting

```
includeFile = includeFile.getCanonicalFile();
```

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,683 @@
...
+        if (!updatedParsedXmls.contains(includeFile)){
+            String onConflict = node.getAttribute("onConflict");
+            if (!"ignore".equalsIgnoreCase(onConflict)){
```
chyt: I mentioned this in a previous comment, but I don't think you addressed it so I'll ask here again: don't you have to first check that there is a conflict? For example, if you set `onConflict="ignore"` doesn't that mean to **ignore if there is a conflict** rather than **always ignore**?

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
chyt: Can you add this under as another info message instead of just a comment in the code?

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,683 @@
...
+     * @throws PluginExecutionException
+     *             if product validation failed or could not be run
+     */
+    public void productInfoValidate() throws PluginExecutionException {
+        Process pr = null;
+        InputStream is = null;
+        Scanner s = null;
+        Worker worker = null;
+        try {
+            String osName = System.getProperty("os.name", "unknown").toLowerCase();
```
chyt: Can `isWindows` be moved to a general util class and made public?

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,683 @@
...
+            worker.start();
+            worker.join(300000);
+            if (worker.exit == null) {
+                throw new PluginExecutionException("Product validation error: timeout");
+            }
+            int exitValue = pr.exitValue();
+            if (exitValue != 0) {
+                is = pr.getInputStream();
+                s = new Scanner(is);
+                s.useDelimiter("\\A");
```
chyt: What is this delimeter?

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,683 @@
...
+            return false;
+        } else {
+            String version1 = extractVersion(file1.getName());
+            String version2 = extractVersion(file2.getName());
+            return compare(version1, version2) < 0;
+        }
+    }
+    
+    private static String extractVersion(String fileName) {
+        int startIndex = INSTALL_MAP_PREFIX.length() + 1;
```
chyt: Why do you add 1 here? If this is to account for the underscore shouldn't that be part of the prefix?

src/test/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtilTest.java
```diff
@@ -0,0 +1,189 @@
...
+        b.add("3");
+        List<String> c = new ArrayList<String>();
+        b.add("4");
+        b.add("5");
+        Set<String> result = InstallFeatureUtil.combineToSet(a, b, c);
+        assertEquals(5, result.size());
+    }
+    
+    @Test
+    public void testProductInfoValidate() throws Exception {
```
chyt: What is the purpose of this test? I think that this should really be an integration test. Also, where are there these files? 

- src/test/resources/installdir/bin/productInfo
- src/test/resources/installdir/bin/productInfo.bat
- src/test/resources/installdir/lib/com.ibm.ws.install.map_1.0.21.jar

They are basically empty. Are they used for this test? If so I think they should be deleted as well.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: Added as another info message.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: Changed to use getCanonicalFile throughout where appropriate.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: Done.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: Done.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: For ignore specifically, the presence of a featureManager section (whether it has features or not) in the parent counts as a conflict, so the child will be ignored.  The absence of a featureManager section counts as no conflict, so the child contents are used.  Verified this against installUtility behaviour and added unit tests for this.  Updated the logic so that a null origResult indicates absence of a featureManager section.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: Moved to OSUtil class.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: Regex to match beginning of input.  Added a comment to clarify.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
ericglau: Yes.  Added a comment to clarify.

src/test/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtilTest.java
ericglau: Removed this test, will add it as an integration test afterwards in maven/gradle.

Removed these files except the map jar (since the InstallFeatureUtil constructor (invoked by other tests) checks for the presence of that file).

ericglau:
@chyt I've addressed all of the remaining comments.  Please review, thanks.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,769 @@
...
+            public boolean accept(File dir, String name) {
+                return name.endsWith(".xml");
+            }
+        });
+        if (configDropinsXmls == null || configDropinsXmls.length == 0) {
+            return null;
+        }
+        Set<String> result = null;
+        for (File xml : configDropinsXmls) {
+            Set<String> features = getServerXmlFeatures(null, xml, null);
```
chyt: For configDropins, I previously asked about the order of the parsing of the files. You said:

> For configDropins, we verified (using installUtility) that features do not get overridden, but are added to the set.

How does this handle `onConflict="replace"`? Is the replace just treated like a merge or something?

ericglau: "replace" replaces all previously parsed features, including any from "configDropins/defaults".  It does not replace features that come after it in the server.xml, and does not replace "configDropins/overrides".  This is tested in the unit test InstallFeatureUtilGetServerFeaturesTest.testCombinationWithReplace().

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -170,89 +172,174 @@ public abstract File downloadArtifact(String groupId, String artifactId, String
...
+     *         files with featureManager sections but no features to install, or
+     *         null if there are no xml files or they have no featureManager
+     *         section
+     */
+    private static Set<String> getConfigDropinsFeatures(File serverDirectory, String folderName) {
+        File configDropinsFolder;
+        try {
+            configDropinsFolder = new File(new File(serverDirectory, "configDropins"), folderName).getCanonicalFile();
+        } catch (IOException e) {
+            // skip this directory if its path cannot be queried
```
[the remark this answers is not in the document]
ericglau: These are in static methods within the abstract class.  The log methods are abstract and implemented by either the Maven or Gradle plugin to use the Maven/Gradle specific logging framework.  I don’t see a way to log from the static method, unless we use something such as log4j or change the method to non-static.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -170,89 +172,174 @@ public abstract File downloadArtifact(String groupId, String artifactId, String
...
-        if (!updatedParsedXmls.contains(includeFile)){
+        if (!updatedParsedXmls.contains(includeFile)) {
             String onConflict = node.getAttribute("onConflict");
-            if (!"ignore".equalsIgnoreCase(onConflict)){
-                if ("".equals(onConflict) || "merge".equalsIgnoreCase(onConflict)){
-                    result.addAll(getServerXmlFeatures(includeFile, updatedParsedXmls));
-                } else if ("replace".equalsIgnoreCase(onConflict)){
-                    result = getServerXmlFeatures(includeFile, updatedParsedXmls);
+            if ("".equals(onConflict) || "merge".equalsIgnoreCase(onConflict)) {
+                Set<String> features = getServerXmlFeatures(null, includeFile, updatedParsedXmls);
```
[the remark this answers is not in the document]
ericglau: Done.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -170,89 +172,174 @@ public abstract File downloadArtifact(String groupId, String artifactId, String
...
+                    // only replace if the child has features
+                    result = features;
                 }
+            } else if ("ignore".equalsIgnoreCase(onConflict)) {
+                Set<String> features = getServerXmlFeatures(null, includeFile, updatedParsedXmls);
+                if (result == null) {
+                    // parent has no results (i.e. no featureManager section), so use the child's results
+                    result = features;
+                } // else the parent already has some results (even if it's empty), so ignore the child
             }
```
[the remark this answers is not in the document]
ericglau: Used the "merge" case for `else`.  Invalid `onConflict` value is treated as merge.

src/main/java/net/wasdev/wlp/common/plugins/util/OSUtil.java
```diff
@@ -0,0 +1,32 @@
...
+public class OSUtil {
+    
+    /**
+     * Determines if the current OS is a Windows OS.
+     * 
+     * @return true if running on Windows, false otherwise
+     */
+    public static boolean isWindows() {
+        String osName = System.getProperty("os.name", "unknown").toLowerCase();
+        boolean isWindows = osName.indexOf("windows") >= 0;
```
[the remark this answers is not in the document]
ericglau: Done.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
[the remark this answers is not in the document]
ericglau: Changed method to non-static and added logging.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -170,89 +172,174 @@ public abstract File downloadArtifact(String groupId, String artifactId, String
...
+    private static Set<String> getServerXmlFeatures(Set<String> origResult, File serverFile, List<File> parsedXmls) {
+        Set<String> result = origResult;
         List<File> updatedParsedXmls = new ArrayList<File>();
-        updatedParsedXmls.add(serverFile);
-        if (parsedXmls != null){
+        File canonicalServerFile;
+        try {
+            canonicalServerFile = serverFile.getCanonicalFile();
+        } catch (IOException e) {
+            // skip this server.xml if its path cannot be queried
```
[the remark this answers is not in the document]
ericglau: Done.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -170,89 +172,174 @@ public abstract File downloadArtifact(String groupId, String artifactId, String
...
-            includeFile = new File(serverFile.getParentFile().getAbsolutePath(), includeFileName);
+        try {
+            if (!includeFile.isAbsolute()) {
+                includeFile = new File(serverFile.getParentFile().getAbsolutePath(), includeFileName)
+                        .getCanonicalFile();
+            } else {
+                includeFile = includeFile.getCanonicalFile();
+            }
+        } catch (IOException e) {
+            // skip this xml if its path cannot be queried
```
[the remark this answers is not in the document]
ericglau: Done.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -170,89 +172,174 @@ public abstract File downloadArtifact(String groupId, String artifactId, String
...
+     *            The list of XML files that have been parsed so far.
+     * @return list of features that should be installed according to the
+     *         origResult and the current serverFile, or empty set if the file
+     *         (or its children) only has a featureManager section with no
+     *         features, or null if the file (and all of its children) has no
+     *         featureManager section
+     */
+    private static Set<String> getServerXmlFeatures(Set<String> origResult, File serverFile, List<File> parsedXmls) {
+        Set<String> result = origResult;
         List<File> updatedParsedXmls = new ArrayList<File>();
```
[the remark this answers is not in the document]
ericglau: Changed to reuse parsedXmls if not null, otherwise create a new instance.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
chyt: What if there is a replace in `configDropins/overrides`? Will that replace everything previously found in `defaults` and `server.xml`?

ericglau: It should replace everything previously found.  Added logic to handle this case and a corresponding unit test case.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,784 @@
...
+        File[] configDropinsXmls = configDropinsFolder.listFiles(new FilenameFilter() {
+            @Override
+            public boolean accept(File dir, String name) {
+                return name.endsWith(".xml");
+            }
+        });
+        if (configDropinsXmls == null || configDropinsXmls.length == 0) {
+            return result;
+        }
+        // sort the files in alphabetical order so that overrides will happen in the proper order
```
chyt: As per [this documentation](https://www.ibm.com/support/knowledgecenter/SSD28V_9.0.0/com.ibm.websphere.wlp.core.doc/ae/twlp_setup_dropins.html), you will need to evaluate all the filenames as lowercase before sorting alphabetically.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,784 @@
...
+        Comparator<File> comparator = new Comparator<File>() {
+            @Override
+            public int compare(File left, File right) {
+                return left.getAbsolutePath().compareTo(right.getAbsolutePath());
```
chyt: Also I think you should be using `getCanonicalPath` here as well instead of `getAbsolutePath`.

ericglau: configDropinsFolder is already canonical, so the list of files is just the file names under that canonical path.  In this case, we should just compare the file paths directly, and not resolve symbolic links.  Verified this with installUtility as well.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -0,0 +1,784 @@
...
+     *            The server XML file.
+     * @param parsedXmls
+     *            The list of XML files that have been parsed so far.
+     * @return The set of features to install, or empty set if the cumulatively
+     *         parsed xml files only have featureManager sections but no
+     *         features to install, or null if there are no valid xml files or
+     *         they have no featureManager section
+     */
+    private Set<String> getServerXmlFeatures(Set<String> origResult, File serverFile, List<File> parsedXmls) {
+        Set<String> result = origResult;
```
chyt: No point in creating another reference to the feature set. You can just modify the `origResult` directly. I would probably also just rename `origResult` to be `result`.

ericglau: origResult could be null, in which case we need to create a new object.  It would be bad practice to modify the reference in the original parameter.

cherylking: It was never documented that origResult could be null. It appeared that you were modifying origResult since you simply set result = origResult.

ericglau: The method doc comments states that it could be null.

ericglau:
https://github.com/OpenLiberty/open-liberty/issues/4061 has been opened to provide a resolver API that includes the server as well, so that InstallFeatureUtil would no longer need to perform its own parsing of the server xmls.

PR #7, opened 2018-06-28
Enable overriding bundle for use with install map

ericglau:
The install map loads other bundles from the runtime into its classpath.  If a newer version of a runtime bundle is made available in a Maven repository, the plugin should use the one from the repository and make it available to the install map to override the one from the runtime.


chyt:
Can you add some unit tests for for `getNextProductVersion`, as well as any other new method that can be unit tested?

ericglau:
Added unit tests for all of the new methods.

PR #10, opened 2018-07-12
Add JAXB API into class path when installing features

ericglau:
This is to address https://github.com/WASdev/ci.maven/issues/343

This does the following:
* Add a dependency on jaxb-api
* Set the parent classloader for the install map
* https://github.com/OpenLiberty/open-liberty/pull/4242 then sets the parent classloader when loading other jars from the install map
* Override the install map with the updated one from the Maven repository.  This will take effect after the updated install map from https://github.com/OpenLiberty/open-liberty/pull/4242 is published.

### DIFF
--- a/src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
+++ b/src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
@@ -94,7 +94,7 @@ public abstract class InstallFeatureUtil {
         if (installJarFile == null) {
             throw new PluginScenarioException("Install map jar not found.");
         }
-        downloadedJsons = downloadProductJsons(installDirectory);
+        downloadedJsons = downloadProductJsons();
         if (downloadedJsons.isEmpty()) {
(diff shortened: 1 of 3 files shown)


src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -671,26 +678,28 @@ public void installFeatures(boolean isAcceptLicense, List<String> featuresToInst
...
-     * @param type
-     *            the type of the bundle e.g. jar
      * @return a String representing the bundle in filepath;BundleName format
      */
-    public String downloadOverrideBundle(String groupId, String artifactId, String type) throws PluginExecutionException {
-        String result = null;
-        File overrideJar = null;
+    public String getOverrideBundleDescriptor(String groupId, String artifactId) throws PluginExecutionException {
+        File overrideJar = downloadOverrideJar(groupId, artifactId);
+        if (overrideJar != null) {
```
chyt: Add a file exists check.

PR #20, opened 2018-09-14
Automatically accept license if installing only Open or WebSphere Liberty features from Maven Central

ericglau:
For https://github.com/WASdev/ci.maven/issues/381

### DIFF
--- a/src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
+++ b/src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
@@ -571,7 +571,7 @@ public abstract class InstallFeatureUtil {
      * @throws PluginExecutionException if any of the JSONs could not be found
      */
     public static Set<String> getLibertyFeatureSet(Set<File> jsons) throws PluginExecutionException {
-        Set<String> openLibertyFeatures = new HashSet<String>();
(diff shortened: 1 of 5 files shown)


ericglau:
@cherylking Yes, integration tests will be added in ci.maven/ci.gradle

ericglau:
Generalized this to include both Open and WebSphere Liberty features.  The exception would be if any features do not come from the two known groupIds on Maven Central.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -560,6 +563,50 @@ private File downloadEsaArtifact(String mavenCoordinates) throws PluginExecution
...
     
+    /**
+     * Gets the set of all Open and WebSphere Liberty features by scanning the product JSONs.
+     * 
+     * @param jsons The set of product JSON files to scan
+     * @return set of all Open and WebSphere Liberty features
+     * @throws PluginExecutionException if any of the JSONs could not be found
+     */
+    public static Set<String> getLibertyFeatureSet(Set<File> jsons) throws PluginExecutionException {
+        Set<String> openLibertyFeatures = new HashSet<String>();
```
cherylking: This is a nit, but probably should rename variable to libertyFeatures.

PR #31, opened 2019-06-07
added new util for getServerFeatures

kathrynkodama:
**Installing features added to the server.xml**
- matches logic in InstallFeatureUtil for getServerFeatures

**Creating missing directories when dev mode is running**
- java source & java test files: if either directories are created an initial compile is done by calling the maven-compiler-plugin and the app is installed by calling the liberty-maven-plugin.

### DIFF
--- a/src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
+++ b/src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
@@ -163,6 +163,8 @@ public abstract class DevUtil {
      * @param configFile
      */
     public abstract void checkConfigFile(File configFile);
+    
+    public abstract boolean initialCompile(File dir);
 
     private List<String> jvmOptions;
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -383,17 +435,33 @@ protected void deleteFile(File deletedFile, File dir, File targetDir) {
...
+            for (File file : fList) {
+                if (file.isFile() && file.getName().toLowerCase().endsWith(".class")) {
+                   file.delete();
+                   info("Deleted Java class file: " + file);
+                } else if (file.isDirectory()) {
+                    cleanTargetDir(file);
+                }
+            }
+        }
+
```
ericglau: if the dir is empty at the end of this method, it would be good to delete the dir itself

PR #33, opened 2019-06-13
Avoid file not found exception

proubatsis:
Add try/catch to avoid file not found exception when copying file.

### DIFF
--- a/src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
+++ b/src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
@@ -420,8 +420,12 @@ public abstract class DevUtil {
                 fileChanged.getAbsolutePath().indexOf(srcDir.getAbsolutePath()) + srcDir.getAbsolutePath().length());
 
         File targetResource = new File(targetDir.getAbsolutePath() + relPath);
-        FileUtils.copyFile(fileChanged, targetResource);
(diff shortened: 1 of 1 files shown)


src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -420,8 +420,12 @@ public void copyFile(File fileChanged, File srcDir, File targetDir) throws IOExc
                 fileChanged.getAbsolutePath().indexOf(srcDir.getAbsolutePath()) + srcDir.getAbsolutePath().length());
 
         File targetResource = new File(targetDir.getAbsolutePath() + relPath);
-        FileUtils.copyFile(fileChanged, targetResource);
-        info("Copied file: " + fileChanged.getAbsolutePath() + " to: " + targetResource.getAbsolutePath());
+        try {
+            FileUtils.copyFile(fileChanged, targetResource);
+            info("Copied file: " + fileChanged.getAbsolutePath() + " to: " + targetResource.getAbsolutePath());
+        } catch (Exception ex) {
```
ericglau: Also catch FileNotFoundException, in which case just log as debug.
When catching any other Exception, log as error and include the exception message.

proubatsis: @ericglau It now handles FileNotFound exception first.

PR #34, opened 2019-06-14
restart dev mode on unhandled change

kathrynkodama:


src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -330,6 +338,18 @@ public void watchFiles(File buildFile, File outputDirectory, File testOutputDire
                     cleanTargetDir(testOutputDirectory);
                     testSourceDirRegistered = false;
                 }
+                
+                // check if defaultConfigDirectory has been added and restart dev mode if it has
+                if (noConfigDir && this.defaultConfigDirectory.exists()){
+                    noConfigDir = false;
```
ericglau: is there any need to set this, since dev mode will be restarted anyways?

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -330,6 +338,18 @@ public void watchFiles(File buildFile, File outputDirectory, File testOutputDire
...
+                    restartDevMode(executor);
+                }
+                
+                // check if configDirectory has been added
+                if (!configDirRegistered && this.configDirectory.exists()){
+                    configDirRegistered = true;
```
ericglau: is there any need to set this, since dev mode will be restarted anyways?

PR #35, opened 2019-06-18
Dev mode debug task

proubatsis:
* Change `enableServerDebug` to create/modify `server.env` file instead of jvm options
* Add `server.env` cleanup to shutdown hook


src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -226,6 +228,27 @@ public void cleanUpJVMOptions() {
             }
         }
     }
+    
+    public void cleanUpServerEnv() {
```
ericglau: reformat this method to use spaces instead of tabs

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -246,34 +270,42 @@ public void run() {
...
-            sb.append("-Dsun.reflect.noInflation=true\n");
-            sb.append("-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=" + libertyDebugPort);
-            BufferedWriter writer = new BufferedWriter(new FileWriter(jvmOptionsFile));
-            writer.write(sb.toString());
-            writer.close();
-            if (jvmOptionsFile.exists()) {
-                info("Successfully created liberty:dev jvm.options file");
-            }
-        }
+    	String serverEnvPath = serverDirectory.getAbsolutePath() + "/server.env";
```
ericglau: reformat this method to use spaces instead of tabs

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -246,34 +270,42 @@ public void run() {
...
+    		while ((line = reader.readLine()) != null) {
+    			sb.append(line);
+    			sb.append("\n");
+    		}
+    		reader.close();
+    	}
+    	
+    	debug("Creating server.env file: " + serverEnvFile.getAbsolutePath());
+    	sb.append("WLP_DEBUG_SUSPEND=n\n");
+    	sb.append("WLP_DEBUG_ADDRESS=");
```
ericglau: what happens if the server.env already has WLP_DEBUG_ADDRESS defined?  Does the new one take precedence?

proubatsis: Yes, the new one takes precedence @ericglau

PR #42, opened 2019-06-24
added unit tests for hot deployment

kathrynkodama:


src/test/java/net/wasdev/wlp/common/plugins/util/DevUtilTest.java
```diff
@@ -76,12 +85,150 @@ public void testCleanupServerEnvBak() throws Exception {
...
+        File configFile = new File(srcDir, "config.xml");
+        Files.write(configFile.toPath(), "temp".getBytes());
+
+        util.copyFile(configFile, srcDir, targetDir, "server.xml");
+
+        File targetFile = new File(targetDir, "server.xml");
+        assertTrue(targetFile.exists());
+
+        // clean up
+        if (srcDir.exists()) {
```
ericglau: If the test fails before this point, the cleanup won't be run.  It's better to cleanup using in a `finally` block or in the tearDown() method.

src/test/java/net/wasdev/wlp/common/plugins/util/DevUtilTest.java
```diff
@@ -76,12 +85,150 @@ public void testCleanupServerEnvBak() throws Exception {
...
+        assertTrue(tempSrcFile.exists());
+        assertTrue(tempTargetFile.exists());
+
+        util.deleteFile(tempSrcFile, srcDir, targetDir, "server.xml");
+
+        // verify that the target file has been deleted
+        assertFalse(tempTargetFile.exists());
+
+        // clean up
+        if (srcDir.exists()) {
```
ericglau: If the test fails before this point, the cleanup won't be run.  It's better to cleanup using in a `finally` block or in the tearDown() method.

src/test/java/net/wasdev/wlp/common/plugins/util/DevUtilTest.java
```diff
@@ -76,12 +85,150 @@ public void testCleanupServerEnvBak() throws Exception {
...
+        assertFalse(tempClass.exists());
+
+        outputDirectory = Files.createTempDirectory("outputDirectory").toFile();
+        File tempTextFile = new File(outputDirectory, "temp.txt");
+        Files.write(tempTextFile.toPath(), "temp".getBytes());
+
+        assertTrue(tempTextFile.exists());
+
+        util.cleanTargetDir(outputDirectory);
+        assertTrue(outputDirectory.exists());
```
ericglau: was this intended to be `assertTrue(tempTextFile.exists());`?

src/test/java/net/wasdev/wlp/common/plugins/util/DevUtilTest.java
```diff
@@ -76,12 +85,150 @@ public void testCleanupServerEnvBak() throws Exception {
...
+
+        if (outputDirectory.exists()) {
```
ericglau: If the test fails before this point, the cleanup won't be run.  It's better to cleanup using in a `finally` block or in the tearDown() method.

src/test/java/net/wasdev/wlp/common/plugins/util/DevUtilTest.java
```diff
@@ -76,12 +85,150 @@ public void testCleanupServerEnvBak() throws Exception {
...
+        File javaClass = new File(classesDir, "temp.class");
+        Files.write(javaClass.toPath(), "temp".getBytes());
+        
+        assertTrue(javaFile.exists());
+        assertTrue(javaClass.exists());
+        
+        util.deleteJavaFile(javaFile, classesDir, compileSourceRoot);
+        assertFalse(javaClass.exists());
+        
+        if (compileSourceRoot.exists()) {
```
ericglau: If the test fails before this point, the cleanup won't be run.  It's better to cleanup using in a `finally` block or in the tearDown() method.

src/test/java/net/wasdev/wlp/common/plugins/util/DevUtilTest.java
kathrynkodama: No I was testing if the outputDirectory exists as the folder should only be deleted if it is empty. The tempTextFile is not a `.class` file so it should not be deleted and the outputDirectory should still exist.

src/test/java/net/wasdev/wlp/common/plugins/util/DevUtilTest.java
```diff
@@ -76,12 +85,150 @@ public void testCleanupServerEnvBak() throws Exception {
...
+        File outputDirectory = Files.createTempDirectory("outputDirectory").toFile();
+        File tempClass = new File(outputDirectory, "temp.class");
+        Files.write(tempClass.toPath(), "temp".getBytes());
+
+        assertTrue(tempClass.exists());
+
+        util.cleanTargetDir(outputDirectory);
+
+        assertFalse(tempClass.exists());
+
```
ericglau: Is the deletion of an empty outputDirectory expected from the cleanTargetDir method?  If so, the method javadoc could mention that, and assert that it has been deleted here.

PR #48, opened 2019-06-28
Dev mode

ericglau:


src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,956 @@
...
+ * Utility class for dev mode.
+ */
+public abstract class DevUtil {
+
+    /**
+     * Log debug
+     * 
+     * @param msg
+     */
+    public abstract void debug(String msg);
```
chyt: Both `DevUtil` and `InstallFeatureUtil` provide abstract methods for logging different types of messages. We recommend pulling these out into a common logger interface, implementing it as a singleton in both Maven and Gradle, and passing that singleton into these classes.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,956 @@
...
+     * Returns whether debug is enabled by the current logger
+     * 
+     * @return whether debug is enabled
+     */
+    public abstract boolean isDebugEnabled();
+
+    /**
+     * Stop the server
+     */
+    public abstract void stopServer();
```
chyt: Can `stopServer` and `startServer` be implemented using Ant directly at the common level? Based on the implementation in ci.maven, it seems like there are only a few parameters you need to retrieve from the build tool configuration. Those could simply be passed in to these methods as needed. The majority of the implementations for these methods in Maven seem like they would apply to Gradle as well.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,956 @@
...
+
+    /**
+     * Starts the default server
+     */
+    public abstract void startServer();
+
+    /**
+     * Updates artifacts of current project
+     */
+    public abstract void getArtifacts(List<String> artifactPaths);
```
chyt: It seems awkward to get passed in a list and then add to it in this method. Why not just return a `List<String>` instead?

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,956 @@
...
+     * @return true if the build file was recompiled with changes
+     */
+    public abstract boolean recompileBuildFile(File buildFile, List<String> artifactPaths, ThreadPoolExecutor executor);
+
+    /**
+     * Get the number of times the application updated message has appeared in the application log
+     * 
+     * @return 
+     */
+    public abstract int countApplicationUpdatedMessages();
```
chyt: If moving start and stop to common, I would also expect this method to be implemented in common as well, with the necessary parameters such as the artifact ID and log file passed in.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,956 @@
...
+
+    /**
+     * Run unit and/or integration tests
+     * 
+     * @param waitForApplicationUpdate Whether to wait for the application to update before running integration tests
+     * @param messageOccurrences The previous number of times the application updated message has appeared.
+     * @param executor The thread pool executor
+     * @param forceSkipUTs Whether to force skip the unit tests
+     */
+    public abstract void runTests(boolean waitForApplicationUpdate, int messageOccurrences, ThreadPoolExecutor executor, boolean forceSkipUTs);
```
chyt: Could this method implementation also be moved to common?

src/main/java/net/wasdev/wlp/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -0,0 +1,314 @@
...
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+import org.xml.sax.SAXException;
+
+import net.wasdev.wlp.common.plugins.config.XmlDocument;
+
+/**
+ * Utility class to determine server features
+ */
+public abstract class ServerFeatureUtil {
```
chyt: If we were to extract the common logger methods as I previously suggested and move them into their own class, could we make the rest of the methods in this class static? It doesn't seem like there is any state being persisted at the class level. This would make the invocations easier in Maven; for example, you could just do `ServerFeatureUtil.getServerFeatures(serverDirectory)`.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,956 @@
...
+                    info("The server configuration file " + configFile + " has been added. Restart liberty:dev mode for it to take effect.");
+                }
+                
+                try {
+                    final WatchKey wk = watcher.poll(1, TimeUnit.SECONDS);
+                    for (WatchEvent<?> event : wk.pollEvents()) {
+                        final Path changed = (Path) event.context();
+
+                        final Watchable watchable = wk.watchable();
+                        final Path directory = (Path) watchable;
```
chyt: Is there any concern that will not be the canonical path?

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,956 @@
...
+                        for (File resourceDir : resourceDirs) {
+                            if (directory.startsWith(resourceDir.toPath())) {
+                                resourceParent = resourceDir;
+                            }
+                        }
+                        
+                        int numApplicationUpdatedMessages = countApplicationUpdatedMessages();
+
+                        // src/main/java directory
+                        if (directory.startsWith(this.sourceDirectory.toPath())) {
```
chyt: When doing these string comparisons here, I think it's more robust if we get the canonical path of the sourceDirectory, etc.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,956 @@
...
+
+    /**
+     * Register the parent directory and all sub-directories with the WatchService
+     * 
+     * @param start parent directory
+     * @param dir path of parent directory
+     * @param watcher WatchService
+     * @throws IOException unable to walk through file tree 
+     */
+    protected void registerAll(final Path start, final Path dir, final WatchService watcher) throws IOException {
```
chyt: Is this `dir` variable used? Seems like it is overwritten on line 693.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,974 @@
...
+                error("Could not move existing liberty:dev server.env file");
+            }
+
+            BufferedReader reader = new BufferedReader(new FileReader(serverEnvBackup));
+            String line;
+            while ((line = reader.readLine()) != null) {
+                sb.append(line);
+                sb.append("\n");
+            }
+            reader.close();
```
cherylking: Anywhere you are using BufferedReader/BufferedWriter or any input stream or output stream, you should wrap the use of it in a try/finally and call the close() method in the finally. This will ensure they do not get left open.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: It is not being overwritten. The `dir` value passed to the registerAll is the same value printed in `debug("Watching directory: " + dir.toString());`

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,974 @@
...
+                        debug("Processing events for watched directory: " + directory);
+
+                        File fileChanged = new File(directory.toString(), changed.toString());
+                        debug("Changed: " + changed + "; " + event.kind());
+
+                        // resource file check
+                        File resourceParent = null;
+                        for (File resourceDir : resourceDirs) {
+                            if (directory.startsWith(resourceDir.toPath())) {
+                                resourceParent = resourceDir;
```
cherylking: Should this break out of the for loop once you find a match?

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,974 @@
...
+                                    && (event.kind() == StandardWatchEventKinds.ENTRY_MODIFY
+                                            || event.kind() == StandardWatchEventKinds.ENTRY_CREATE)) {
+                                debug("Java source file modified: " + fileChanged.getName());
+
+                                // tests are run in recompileJavaSource
+                                recompileJavaSource(javaFilesChanged, artifactPaths, executor, outputDirectory,
+                                        testOutputDirectory);
+                            } else if (event.kind() == StandardWatchEventKinds.ENTRY_DELETE) {
+                                debug("Java file deleted: " + fileChanged.getName());
+                                deleteJavaFile(fileChanged, outputDirectory, this.sourceDirectory);
```
cherylking: Does this recompile the Java source? After deleting a Java file, it probably should.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,974 @@
...
+            String fileName = fileChanged.getName().substring(0, fileChanged.getName().indexOf(".java"));
+            File parentFile = fileChanged.getParentFile();
+            String relPath = parentFile.getCanonicalPath()
+                    .substring(parentFile.getCanonicalPath().indexOf(compileSourceRoot.getCanonicalPath())
+                            + compileSourceRoot.getCanonicalPath().length())
+                    + "/" + fileName + ".class";
+            File targetFile = new File(classesDir.getCanonicalPath() + relPath);
+
+            if (targetFile.exists()) {
+                targetFile.delete();
```
cherylking: Probably should check the return boolean to see if the file was deleted successfully and print info accordingly.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -0,0 +1,974 @@
...
+        Set<File> classPathElements = new HashSet<>();
+        classPathElements.addAll(outputDirs);
+        while (!toParse.isEmpty()) {
+            String s = toParse.poll();
+            if (!parsedFiles.contains(s)) {
+                parsedFiles.add(s);
+                File file = new File(s);
+                if (file.exists() && file.getName().endsWith(".jar")) {
+                    classPathElements.add(file);
+                    if (!file.isDirectory() && file.getName().endsWith(".jar")) {
```
cherylking: You already checked that it ends with ".jar" at line 897. This check here seems unnecessary.

src/main/java/net/wasdev/wlp/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -0,0 +1,314 @@
+package net.wasdev.wlp.common.plugins.util;
```
cherylking: Add copyright

src/main/java/net/wasdev/wlp/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -0,0 +1,314 @@
...
+     *            The list of XML files that have been parsed so far.
+     * @return The set of features to install, or empty set if the cumulatively
+     *         parsed xml files only have featureManager sections but no
+     *         features to install, or null if there are no valid xml files or
+     *         they have no featureManager section
+     */
+    private Set<String> parseIncludeNode(Set<String> origResult, File serverFile, Element node,
+            List<File> updatedParsedXmls) {
+        Set<String> result = origResult;
+        String includeFileName = node.getAttribute("location");
```
cherylking: If the location attribute contains a reference to a variable, what happens here? Is that allowed? Not sure.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: fixed with https://github.com/WASdev/ci.common/pull/48/commits/83e2391e82f984eb04367b51310869a7c9853607

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: fixed with https://github.com/WASdev/ci.common/pull/48/commits/83e2391e82f984eb04367b51310869a7c9853607

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: As this is not specific to just Dev Mode but InstallFeatureUtil as well, this will be tracked as a new issue: https://github.com/WASdev/ci.common/issues/51

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: This does not seem specific to Dev Mode as other mojos (StartServer, StopServer, etc.) execute Ant tasks in ci.maven. If we were to implement `stopServer` and `startServer` for Dev Mode in ci.common it would make sense to make this change for every `stopServer` and `startServer` in ci.maven.  If appropriate we can open an issue to track making this change for all of ci.maven.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
ericglau: In general, we only recompile the java files that were changed.  In this case, a java file was deleted so we just delete the corresponding class file.  We could potentially recompile the entire source in order to surface any compile errors if other classes reference the deleted class, but that seems excessive.  In the current implementation, what someone would see (if something still referenced the deleted file) would be an error at runtime, as well as a compile error shown by their language server in whatever IDE they are using (which should be enough for this scenario).

src/main/java/net/wasdev/wlp/common/plugins/util/ServerFeatureUtil.java
ericglau: This class was just refactored from the existing InstallFeatureUtil.  Regardless, the location attribute is in server.xml and it cannot have a reference to a variable because it needs to be able to work outside of Maven (e.g. it should work with the installUtility tool if using WebSphere Liberty)

src/main/java/net/wasdev/wlp/common/plugins/util/ServerFeatureUtil.java
kathrynkodama: fixed in https://github.com/WASdev/ci.common/pull/52

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: fixed in https://github.com/WASdev/ci.common/pull/52

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: fixed in https://github.com/WASdev/ci.common/pull/52

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: fixed in https://github.com/WASdev/ci.common/pull/52

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: fixed in https://github.com/WASdev/ci.common/pull/52

src/main/java/net/wasdev/wlp/common/plugins/util/ServerFeatureUtil.java
cherylking: I just verified that the location attribute can have a variable in it (not a maven variable..that is not what I meant....a variable defined in a Liberty file (like server.env, another include file or server.xml) or an environment variable. https://www.ibm.com/support/knowledgecenter/en/SSEQTP_liberty/com.ibm.websphere.wlp.doc/ae/twlp_setup_includes.html

ericglau: Opened a separate issue to track this as it affects install feature as well.  https://github.com/WASdev/ci.common/issues/53

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: This `java.nio.file.Path` object used here will be a relative path to the watched directory `typically represent a system dependent file path`.  It does not need to be the canonical path in this case.  We then calculate the fileChanged path on line 481: 
```java
File fileChanged = new File(directory.toString(), changed.toString());
```

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: Addressed above, we can open an issue to track using Ant directly at the common level.  Until that issue is resolved I think it is best to leave this as an abstract method for now.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
kathrynkodama: Addressed above, we can open an issue to track using Ant directly at the common level. Until that issue is resolved I think it is best to leave this as an abstract method for now.

src/main/java/net/wasdev/wlp/common/plugins/util/ServerFeatureUtil.java
kathrynkodama: Addressed above, extracting the common logger methods this will be tracked as a new issue: #51.  Once that issue is addressed we could then make the rest of the methods static.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
chyt: The reason I suggest this is that the start server implementation for dev mode is much more involved than the existing implementations in `StartServerMojo` and `StopServerMojo` - particularly in setting up the WatchService for the logs. We can look into refactoring those into common, but for the most part they are just wrapping the ant task and performing little additional behavior.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
chyt: Why was this issue marked resolved without comment?

ericglau: it was changed as suggested

chyt: If it was changed wouldn't this comment get marked as outdated? I am still seeing it returning a `void` instead of `List<String>`.

ericglau: Not sure why.  The updated version can be seen here https://github.com/WASdev/ci.common/pull/48/files#diff-40ef4b1443d5315fd48803cbbcb671b0R137

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
ericglau: This would involve refactoring most of `StartDebugMojoSupport` into ci.common, which I think should be handled in a separate issue.

ericglau: Opened https://github.com/WASdev/ci.common/issues/54

chyt: I did not intend to suggest refactoring what is in `StartDebugMojoSupport`. Rather, I would like to see what is in that method implementation in `DevMojo` in Maven moved to common. So initialize the `ServerTask` in common, and pass in whatever parameters that are needed from Maven/Gradle (`serverName`, `outputDir`) into `startServer()` to perform the initialization. Aside from the task initialization, I think the rest of what is in that method in Maven would apply to Gradle as well, right?

ericglau: Thanks, I see what you mean.  I've refactored `startServer` into ci.common in these commits: 
https://github.com/WASdev/ci.maven/pull/476/commits/7ec10c9256d90943e48b13e898965bd1b9252e63 
https://github.com/WASdev/ci.common/pull/48/commits/e5747ef4b3f812dd3640c35897642bd7c88bbd08 
https://github.com/WASdev/ci.common/pull/48/commits/b8149f14aca5d3eb20a251a18712a622fb79e897

Didn't refactor `stopServer` since that's just a few lines and would have no benefit because it still needs to call `initializeJava` in ci.maven.

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
ericglau: Moved to common with the following commits:
https://github.com/WASdev/ci.common/pull/48/commits/23d18416f984b2112956ab39f66982ec1925e752
https://github.com/WASdev/ci.maven/pull/476/commits/fcc8ea545cc8a419be9877be989f653a84725799

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
ericglau: Moved to common in the following commits:
https://github.com/WASdev/ci.common/pull/48/commits/72a6d9ff57e101f0d34dd3771a49e03947bd9f32
https://github.com/WASdev/ci.maven/pull/476/commits/b51725037b22f1cc90ed46f11e1ec155966d1d6d

PR #49, opened 2019-07-09
converting absolute paths to canonical paths

kathrynkodama:


src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -718,20 +723,21 @@ protected File getFileFromConfigDirectory(String file) {
      * @param fileChanged Java file changed
      * @param classesDir the directory for compiled classes
      * @param compileSourceRoot the source directory for the Java classes
+     * @throws IOException
```
ericglau: add javadoc comment on the conditions that can cause IOException

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -841,8 +847,9 @@ protected void recompileJava(List<File> javaFilesChanged, List<String> artifactP
      * @param artifactPaths list of artifacts for the current project
      * @param outputDirs list of output directories for the current project
      * @return set of classpath files
+     * @throws IOException
```
ericglau: add javadoc comment on the conditions that can cause IOException

PR #58, opened 2019-07-29
More reliabily read output from productInfo validate

chyt:
Resolves an issue with the installation of some features, where the call to `productInfo validate` was hanging and timing out.

### DIFF
--- a/src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
+++ b/src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
@@ -709,13 +709,14 @@ public abstract class InstallFeatureUtil extends ServerFeatureUtil {
     
     /**
      * Runs the productInfo command and returns the output
+     * Made public static for tests to use in LMP/LGP
      * 
      * @param installDirectory The directory of the installed runtime
      * @param action           The action to perform for the productInfo command
      * @return The command output
      * @throws PluginExecutionException if the exit value of the command was not 0
      */
-    private String productInfo(File installDirectory, String action) throws PluginExecutionException {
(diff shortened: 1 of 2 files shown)


src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -709,55 +711,49 @@ private void productInfoValidate() throws PluginExecutionException {
...
-            worker.start();
-            worker.join(300000);
-            if (worker.exit == null) {
+            ProcessBuilder pb = new ProcessBuilder(productInfoFile, action);
+            pr = pb.start();
+
+            in = new BufferedReader(new InputStreamReader(pr.getInputStream()));
+            String line;
+            while ((line = in.readLine()) != null) {
+                sb.append(line);
```
ericglau: Use `sb.append(line).append(System.lineSeparator());` so that the result includes line breaks, since it will be displayed to the user if there is a validation error.

src/main/java/net/wasdev/wlp/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -709,55 +711,49 @@ private void productInfoValidate() throws PluginExecutionException {
      * Runs the productInfo command and returns the output
      * 
      * @param installDirectory The directory of the installed runtime
-     * @param action The action to perform for the productInfo command
+     * @param action           The action to perform for the productInfo command
      * @return The command output
      * @throws PluginExecutionException if the exit value of the command was not 0
      */
-    public static String productInfo(File installDirectory, String action) throws PluginExecutionException {
```
ericglau: Use `public static` since this method is called from ci.maven's integration tests.

PR #60, opened 2019-08-16
stop dev mode on server stop

kathrynkodama:
Handles `liberty:stop-server` while dev mode is running and `kill $pid` of the dev mode process.


src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -680,6 +681,10 @@ public void watchFiles(File buildFile, File outputDirectory, File testOutputDire
             debug("Watching build file directory: " + buildFile.getParentFile().toPath());
 
             while (true) {
+                if (serverThread.getState().equals(Thread.State.TERMINATED)){
+                    System.exit(0);
```
chyt: Is there a way to terminate more gracefully? Perhaps with a `PluginScenarioException`?

PR #61, opened 2019-08-19
stop dev mode if there is a port conflict

kathrynkodama:


src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -417,6 +418,13 @@ public void run() {
             long timeout = verifyTimeout * 1000;
             long endTime = System.currentTimeMillis() + timeout;
 
+            // Check for port already in use error
+            String portError = serverTask.waitForStringInLog(PORT_IN_USE_MESSAGE_REGEXP, timeout, messagesLogFile);
```
ericglau: If the port is not actually in use, would this cause it to wait the entire timeout duration before it continues to check whether the server is started?

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -417,6 +418,13 @@ public void run() {
...
+            if (portError != null) {
+            	stopServer();
+            	throw new PluginExecutionException("Configured port may already be in use.");
```
ericglau: Does it reach this message after calling stopServer() above?  Also improve the message to provide more details such as port number, or just reuse the error message retrieved from the log.

kathrynkodama: Modified to reuse the error message retrieved from the log, it does reach the message after the stopServer(). 

However, occasionally it will still display `The server has stopped. Exiting dev mode` error from https://github.com/OpenLiberty/ci.common/blob/9cca388c94b432870873237b78138b5ff56846e0/src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java#L695

I will open a separate issue for this.

ericglau: Sure. Sounds like a timing issue.

PR #62, opened 2019-08-20
do not throw stop server exception if dev mode stops the server itself

kathrynkodama:
Track with a flag whether dev mode stopped the server. If `devStop` is `true` dev mode stopped the server, if `devStop` is `false` an external process stopped the server.


src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -211,6 +211,7 @@
     private String applicationId;
     private int appUpdateTimeout;
     private Thread serverThread;
+    private boolean devStop;
```
ericglau: Since multiple threads (the main DevUtil thread and the shutdown hook thread) can modify this value, we should use something like AtomicBoolean.

PR #63, opened 2019-08-21
Parse hostname and ports on dev mode server startup

ericglau:


src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -428,18 +433,99 @@ public void run() {
...
+        for (int i = messageTokens.length - 1; i >= 0; i--) {
+            String numericToken = messageTokens[i].replaceAll("[^\\d]", "" );
+            try {
+                int parsedPort = Integer.parseInt(numericToken);
+                if (parsedPort <= 65535) {
+                    port = numericToken;
+                    break;
+                }
+            } catch (NumberFormatException e) {
+                // ignore
```
kathrynkodama: Should we log a debug message if this exception occurs?

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -428,18 +433,99 @@ public void run() {
...
+
+    private String getPortFromMessageTokens(String[] messageTokens) {
+        String port = null;
+        // For each space-separated token, keep only the numeric parts.
+        // The port is the last numeric token which is a number <= 65535.
+        for (int i = messageTokens.length - 1; i >= 0; i--) {
+            String numericToken = messageTokens[i].replaceAll("[^\\d]", "" );
+            try {
+                int parsedPort = Integer.parseInt(numericToken);
+                if (parsedPort <= 65535) {
```
kathrynkodama: If we check for an upper bound of the port number should we check for a lower bound? ie. 1024

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
ericglau: Added debug message

src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
ericglau: There is no lower bound.  User could possibly set the port to one of the well-known ports.

PR #71, opened 2019-08-28
added run test message after every test run and scanner.nextLine check

kathrynkodama:


src/main/java/net/wasdev/wlp/common/plugins/util/DevUtil.java
```diff
@@ -720,13 +719,18 @@ public void shutdown() {
...
                 debug("Waiting for Enter key to run tests");
-                String line = scanner.nextLine();
-                if (line != null && line.trim().equalsIgnoreCase("exit")) {
-                    debug("Detected exit command");
-                    System.exit(0);
+                if (scanner.hasNextLine()) {
+                    String line = scanner.nextLine();
+                    if (line != null && line.trim().equalsIgnoreCase("exit")) {
+                        debug("Detected exit command");
+                        System.exit(0);
```
chyt: Noticed we're still doing a `System.exit()` here, I recall we discussed throwing an exception instead, could we do that here as well?

kathrynkodama: I do not think we want to throw an exception here. This is the case that the user enters "exit" to stop dev mode (instead of ctl-c). This is expected behaviour so we do not want the user to see an error. The system.exit(0) enters the shutdown hook to properly shut down dev mode.

chyt: What does this look like to the user? Do they still receive some build success output at the end?

kathrynkodama: Yes they do

PR #76, opened 2019-09-04
Add a compile wait time before processing Java changes

ericglau:
Wait at least a certain amount of time since the last detected Java change before processing compilation or deletions.  This is to allow time for changes to be saved when refactoring multiple files, so that all files can be compiled at the same time to avoid compilation errors.

Fix for https://github.com/OpenLiberty/ci.maven/issues/533


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -793,13 +797,69 @@ public void watchFiles(File buildFile, File outputDirectory, File testOutputDire
...
+
             while (true) {
 
                 // stop dev mode if the server has been stopped by another process
                 if (serverThread.getState().equals(Thread.State.TERMINATED) && (this.devStop.get() == false)) {
                     throw new PluginScenarioException("The server has stopped. Exiting dev mode.");
                 }
 
+                // process java source files if no changes detected after the compile wait time
+                boolean processSources = System.currentTimeMillis() > lastJavaSourceChange + compileWaitMillis;
```
kathrynkodama: should this call `this.compileWaitMillis`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1220,8 +1275,16 @@ protected void recompileJava(List<File> javaFilesChanged, List<String> artifactP
...
-            Iterable<? extends JavaFileObject> compilationUnits = fileManager
-                    .getJavaFileObjectsFromFiles(javaFilesChanged);
+            Collection<JavaFileObject> compilationUnits = new HashSet<JavaFileObject>();
+            for (File file : javaFilesChanged) {
+                if (file.exists()) {
+                    for (JavaFileObject o : fileManager.getJavaFileObjects(file)) {
+                        compilationUnits.add(o);
+                    }    
+                } else {
+                    debug("The java file " + file + " does not exist and will not be compiled.");
```
kathrynkodama: The word `java` should be capitalized.

ericglau: fixed

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: It doesn't need `this.` since the variable is not ambiguous in this scope.

PR #86, opened 2019-09-20
detect server status when it is killed

patricktiu:
Signed-off-by: Patrick Tiu <tiu@ca.ibm.com>


src/main/java/io/openliberty/tools/common/plugins/util/ServerStatusUtil.java
```diff
@@ -54,12 +56,20 @@ public static boolean isServerRunning(File outputDirectory, String serverName) {
...
+                try {
+                    String env[] = { "WLP_OUTPUT_DIR=" + outputDirectory };
+                    String cmd[] = { serverStatusCmd, "status", serverName };
+                    Process p = Runtime.getRuntime().exec(cmd, env);
+                    p.waitFor(10, TimeUnit.SECONDS);
+                    if (p.exitValue() != 0) {
+                        return false;
+                    }
+                } catch (Exception e) {
+                    e.printStackTrace();
```
ericglau: It should not print stacktrace here but handle the exception as appropriate.

PR #94, opened 2019-09-30
redeploy app when classes dir is created

kathrynkodama:
partial fix for [issue 569](https://github.com/OpenLiberty/ci.maven/issues/569)

Redeploy application if the classes directory must be created (first java class file is created).

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -54,6 +54,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Scanner;
 import java.util.Set;
 import java.util.concurrent.RejectedExecutionException;
@@ -1485,7 +1486,7 @@ public abstract class DevUtil {
                 if (!classesDir.mkdirs()) {
                     throw new PluginExecutionException("The classes output directory " + classesDir.getAbsolutePath()
                             + " does not exist and cannot be created.");
-                } else {
+                } else if (classesDir.exists() && Objects.equals(classesDir.getCanonicalFile(), outputDirectory.getCanonicalFile())) {
                     // redeploy application when class directory has been created
                     redeployApp();
                 }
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1416,9 +1421,14 @@ protected void recompileJava(Collection<File> javaFilesChanged, List<String> art
...
-
-            if (!classesDir.exists() && !classesDir.mkdirs()) {
-                throw new PluginExecutionException("The classes output directory " + classesDir.getAbsolutePath() + " does not exist and cannot be created.");
+            if (!classesDir.exists()) {
+                if (!classesDir.mkdirs()) {
+                    throw new PluginExecutionException("The classes output directory " + classesDir.getAbsolutePath()
+                            + " does not exist and cannot be created.");
+                } else {
+                    // redeploy application when class directory has been created
+                    redeployApp();
```
patricktiu: redeploy the app only if class dir is created (i.e. outputDirectory) and we don't need to redeploy the app after the first test class is created in testOutputDirectory.

kathrynkodama: Addressed in commit https://github.com/OpenLiberty/ci.common/pull/94/commits/f297b3ddc49faba28d1395a59020ca206bdfa8f9

src/test/java/io/openliberty/tools/common/plugins/util/BaseDevUtilTest.java
```diff
@@ -147,6 +147,11 @@ public void runUnitTests() throws PluginScenarioException, PluginExecutionExcept
         public void runIntegrationTests() throws PluginScenarioException, PluginExecutionException {
             // not needed for tests
         }
+
+        @Override
+        public void redeployApp() throws Exception {
```
patricktiu: should we throws explicit exception, such as PluginExecutionException, so that we can catch it in the recompileJava() method and output the meaning error message.

kathrynkodama: Fixed with https://github.com/OpenLiberty/ci.common/pull/94/commits/3d1a1238bd465a59602014dc4a8f51e27c5bd0f2

PR #99, opened 2019-10-03
Finding app name defined in server.xml using app location

mattbsox:


src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -240,17 +243,26 @@ private static void parseNames(Document doc, String expression) throws XPathExpr
...
+                            names.add(nameValue);
+                            if (resolvedLocation != null) {
+                                locationsAndNames.put(resolvedLocation, nameValue);
+                            }
+                        }
+                    } else if (!names.contains(resolvedName)) {
+                        names.add(resolvedName);
+                        
+                        if (resolvedLocation != null) {
+                            locationsAndNames.put(resolvedLocation, resolvedName);
```
cherylking: I think perhaps an app could have the same name in multiple locations? I would suggest pulling line 257 and 264 out of the if/else if/else if statement. Put it right after between 266 and 267. Then you will get all locations mapped to their app names.

cherylking: One more point...you need to reinitialize the HashMap in initializeAppsLocation.

PR #102, opened 2019-10-21
do not throw runtime exception if dev stopped with ctl-c

kathrynkodama:
Fixes https://github.com/OpenLiberty/ci.maven/issues/661

If dev mode is stopped with Ctl-c, do not throw the server start error.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -445,9 +445,13 @@ public abstract class DevUtil {
                     try {
                         serverTask.execute();
                     } catch (RuntimeException e) {
-                        // If a runtime exception occurred in the server task, log and rethrow
-                        error("An error occurred while starting the server: " + e.getMessage(), e);
-                        throw e;
+                        // If devStop is true server was stopped with Ctl-c, do not throw exception
+                        AtomicBoolean devStop = getDevStop();
+                        if (devStop.get() == false) {
+                            // If a runtime exception occurred in the server task, log and rethrow
+                            error("An error occurred while starting the server: " + e.getMessage(), e);
+                            throw e;
+                        }
                     }
                 }
 
@@ -624,6 +628,10 @@ public abstract class DevUtil {
     
     public void setDevStop(boolean devStop) {
         this.devStop.set(devStop);
+    }
+
+    public AtomicBoolean getDevStop() {
+        return this.devStop;
     }
 
     public void addShutdownHook(final ThreadPoolExecutor executor) {


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -445,9 +445,13 @@ public void run() {
                     try {
                         serverTask.execute();
                     } catch (RuntimeException e) {
-                        // If a runtime exception occurred in the server task, log and rethrow
-                        error("An error occurred while starting the server: " + e.getMessage(), e);
-                        throw e;
+                        // If devStop is true server was stopped with Ctl-c, do not throw exception
+                        AtomicBoolean devStop = getDevStop();
```
ericglau: The `getDevStop()` method shouldn't be necessary since `devStop` is an instance variable in DevUtil.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -445,9 +445,13 @@ public void run() {
...
+                        if (devStop.get() == false) {
```
ericglau: In the `addShutdownHook` method where it has
```
                cleanUpTempConfig();
                cleanUpServerEnv();
                setDevStop(true);
```
, we should move setDevStop to be first thing that happens, otherwise there may be some race condition here.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
kathrynkodama: Since the error in caught inside the `new Runnable() {` object we cannot access `this.devStop` directly.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
kathrynkodama: Addressed with https://github.com/OpenLiberty/ci.common/pull/102/commits/647948f0a2d2638e8766a646768e7cfae15c641d

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: I don't see any compile error if I remove the line `AtomicBoolean devStop = getDevStop();`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: I just realized it would have worked both ways since the shutdown hook doesn't stop the server until after the above mentioned part.  But the new way is neater.

PR #106, opened 2019-10-28
Calculate logDirectory using serverDirectory

rzgry:


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -437,7 +437,7 @@ public void startServer() throws PluginExecutionException {
             // Set debug variables in server.env if debug enabled
             enableServerDebug();
 
-            String logsDirectory = serverTask.getOutputDir() + "/" + serverTask.getServerName() + "/logs";
+            String logsDirectory = serverDirectory.getCanonicalPath() + "/" + "/logs";
```
ericglau: Remove the extra `"/"`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -356,8 +356,7 @@ public void runTests(boolean waitForApplicationUpdate, int messageOccurrences, T
                     }
                     long timeout = appStartupTimeout * 1000;
 
-                    String logsDirectory = serverTask.getOutputDir() + "/" + serverTask.getServerName() + "/logs";
-                    File messagesLogFile = new File(logsDirectory + "/messages.log");
+                    File messagesLogFile = serverTask.getLogFile();
```
ericglau: This line could be removed, and just use the `logFile` variable instead.

PR #107, opened 2019-10-30
support for deleting java src directories or config directories

kathrynkodama:
Fixes https://github.com/OpenLiberty/ci.maven/issues/622

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1340,10 +1340,15 @@ public abstract class DevUtil {
         File targetFile = getTargetFile(deletedFile, dir, targetDir, targetFileName);
         debug("Target file exists: " + targetFile.exists());
         if (targetFile.exists()) {
-            if (targetFile.delete()){
-                info("Deleted file" + targetFile.getCanonicalPath());
+            if (targetFile.isDirectory()) {
+                FileUtils.deleteDirectory(targetFile);
+                info("Directory deleted " + targetFile.getCanonicalPath());
             } else {
-                error("Error deleting file " + targetFile.getCanonicalPath());
+                if (targetFile.delete()){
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1340,10 +1340,21 @@ protected void deleteFile(File deletedFile, File dir, File targetDir, String tar
...
-                info("Deleted file" + targetFile.getCanonicalPath());
+            if (targetFile.isDirectory()) {
+                try {
+                    FileUtils.deleteDirectory(targetFile);
+                } catch (IllegalArgumentException e) {
+                    debug("Could not delete directory " + e.getMessage());
+                } catch (IOException e) {
+                    error("Error deleting directory " + e.getMessage());
+                }
+                info("Directory deleted " + targetFile.getCanonicalPath());
```
ericglau: This info message should move within the `try` statement, otherwise it would print even if the directory wasn't deleted.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1340,10 +1340,21 @@ protected void deleteFile(File deletedFile, File dir, File targetDir, String tar
         File targetFile = getTargetFile(deletedFile, dir, targetDir, targetFileName);
         debug("Target file exists: " + targetFile.exists());
         if (targetFile.exists()) {
-            if (targetFile.delete()){
-                info("Deleted file" + targetFile.getCanonicalPath());
+            if (targetFile.isDirectory()) {
+                try {
+                    FileUtils.deleteDirectory(targetFile);
+                } catch (IllegalArgumentException e) {
+                    debug("Could not delete directory " + e.getMessage());
```
ericglau: Use sentences for the new debug and error messages (lines 1347, 1349, 1458, 1460), e.g. they are missing a period.

PR #118, opened 2019-12-09
warn user to restart if resource dir is deleted

kathrynkodama:
Decided to warn user to restart dev mode when a resource directory is deleted.  Tracking the files within each resource directory to know which to delete upon deletion of the directory is too expensive.

Fixes https://github.com/OpenLiberty/ci.maven/issues/700

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1136,13 +1136,13 @@ public abstract class DevUtil {
                 }
                 
                 // check if resourceDirectory has been added
-                for (File resourceDir : resourceDirs){
-                    if (!resourceMap.get(resourceDir)) {
-                        if (resourceDir.exists()) {
-                            registerAll(resourceDir.getCanonicalFile().toPath(), watcher);
-                            resourceMap.put(resourceDir, true);
-                        }
-                    } else {
+                for (File resourceDir : resourceDirs) {
+                    if (!resourceMap.get(resourceDir) && resourceDir.exists()) {
+                        // added resource directory
+                        registerAll(resourceDir.getCanonicalFile().toPath(), watcher);
+                        resourceMap.put(resourceDir, true);
+                    } else if (resourceMap.get(resourceDir) && !resourceDir.exists()) {
+                        // deleted resource directory
                         warn("The resource directory " + resourceDir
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1142,6 +1142,9 @@ public void watchFiles(File buildFile, File outputDirectory, File testOutputDire
                             registerAll(resourceDir.getCanonicalFile().toPath(), watcher);
                             resourceMap.put(resourceDir, true);
                         }
+                    } else if (resourceMap.get(resourceDir) && !resourceDir.exists()) {
```
ericglau: This doesn't need the `resourceMap.get(resourceDir)` condition since it's implied by the `else` clause.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1142,6 +1142,9 @@ public void watchFiles(File buildFile, File outputDirectory, File testOutputDire
...
+                        warn("The resource directory " + resourceDir + "was deleted.  Restart liberty:dev mode for it to take effect.");
```
ericglau: Missing a space, should be: `resourceDir + " was deleted.`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1142,8 +1142,9 @@ public void watchFiles(File buildFile, File outputDirectory, File testOutputDire
                             registerAll(resourceDir.getCanonicalFile().toPath(), watcher);
                             resourceMap.put(resourceDir, true);
                         }
-                    } else if (resourceMap.get(resourceDir) && !resourceDir.exists()) {
-                        warn("The resource directory " + resourceDir + "was deleted.  Restart liberty:dev mode for it to take effect.");
+                    } else {
```
ericglau: It's missing a condition.
Do one of the following instead.  The first one looks less confusing.

```
if (!resourceMap.get(resourceDir) && resourceDir.exists()) {
    // added
    ...
} else if (resourceMap.get(resourceDir) && !resourceDir.exists()) {
   // deleted
```
or
```
if (!resourceMap.get(resourceDir)) {
    if (resourceDir.exists()) {
        //added
        ...
    }
} else if (!resourceDir.exists()) {
    // deleted
```

PR #125, opened 2020-01-30
Fix variable resolution for variables starting with env.

cherylking:
This is for gradle issue https://github.com/OpenLiberty/ci.gradle/issues/426.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
@@ -484,11 +484,7 @@ public class ServerConfigDocument {
         }
 
         for (String nextVariable : variablesToResolve) {
-            String value = getProperties().getProperty(nextVariable);
-            if (value == null) {
-                // Check for default value since no other value found.
-                value = getDefaultProperties().getProperty(nextVariable);
-            }
+            String value = getPropertyValue(nextVariable);
 
             if (value != null && !value.isEmpty()) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -517,6 +513,24 @@ private static String resolveVariables(String nodeValue, Collection<String> vari
         return resolved;
     }
 
+    private static String getPropertyValue(String propertyName) {
```
chyt: The behavior to resolve the property is the same here regardless of whether it is prefixed by `env.` or not. It would be more concise to check for the prefix first, strip it if necessary, then apply the property resolution.

cherylking: But what if they happened to define a variable that starts with `env.`? I don't think anything prevents that, and that should be resolved before attempting to resolve the variable without the prefix.

chyt: I would expect that if the `env` prefix is reserved as an indicator for environment variables then a user shouldn't be allowed to define non-environment variables that start with that prefix. We should probably find out how this is handled in Liberty so we can ensure consistent behavior.

chyt: Documenting that @cherylking ensured via testing that non-environment variables can use the `env` prefix.

PR #128, opened 2020-02-07
Disable file caching while we load features in Liberty then restore the caching state

turkeylurkey:
… afterwards. This fix is for issue [gradle-402](https://github.com/OpenLiberty/ci.gradle/issues/402)

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
@@ -24,6 +24,7 @@ import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
@@ -66,6 +67,7 @@ public abstract class InstallFeatureUtil extends ServerFeatureUtil {
     private static final String INSTALL_MAP_SUFFIX = ".jar";
     private static final String OPEN_LIBERTY_PRODUCT_ID = "io.openliberty";
     private String openLibertyVersion;
+    private static Boolean saveURLCacheStatus = null;
 
     /**
      * Initialize the utility and check for unsupported scenarios.
@@ -434,6 +436,7 @@ public abstract class InstallFeatureUtil extends ServerFeatureUtil {
         }
         Map<String, Object> mapBasedInstallKernel = null;
 
+        disableCacheInURLClassLoader();
         try (final URLClassLoader loader = new URLClassLoader(new URL[] { installJarURL }, getClass().getClassLoader())) {
             mapBasedInstallKernel = createMapBasedInstallKernelInstance(loader, installDirectory);
             mapBasedInstallKernel.put("install.local.esa", true);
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -507,6 +510,39 @@ public void installFeatures(boolean isAcceptLicense, List<String> featuresToInst
...
+    private synchronized void disableCacheInURLClassLoader() {
+        try {
+            if (saveURLCacheStatus == null) {
+                Method getDefaultCaching = java.net.URLConnection.class.getMethod("getDefaultUseCaches", String.class);
+                saveURLCacheStatus = Boolean.valueOf((boolean) getDefaultCaching.invoke(null, "jar")); // null = static method
+                Method disableCaching = java.net.URLConnection.class.getMethod("setDefaultUseCaches", String.class, boolean.class);
+                disableCaching.invoke(null, "jar", false); // null = static method, false = do not cache
+            }
+        } catch (Exception e) {  // ignore the exceptions if this does not work. E.g. no method found in Java 8.
+            info(e.getClass().getName() + " trying to invoke java.net.URLConnection.setDefaultUseCaches(S,b) in disable");
```
ericglau: - Just catch `NoSuchMethodException` otherwise this will suppress any other possible exceptions.
- Log as debug instead of info, otherwise Java 8 users will always see this message.

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -507,6 +510,39 @@ public void installFeatures(boolean isAcceptLicense, List<String> featuresToInst
...
+    // will be cached or not as previously set. Use reflection because this is not supported
+    // in Java 8.
+    private synchronized void restoreCacheInURLClassLoader() {
+        try {
+            if (saveURLCacheStatus != null) {
+                Method disableCaching = java.net.URLConnection.class.getMethod("setDefaultUseCaches", String.class, boolean.class);
+                disableCaching.invoke(null, "jar", saveURLCacheStatus.booleanValue()); // null = static method
+            }
+        } catch (Exception e) {  // ignore the exceptions if this does not work. E.g. no method found in Java 8.
+            info(e.getClass().getName() + " trying to invoke java.net.URLConnection.setDefaultUseCaches(S,b) in restore");
```
ericglau: - Just catch `NoSuchMethodException` otherwise this will suppress any other possible exceptions.
- Log as debug instead of info, otherwise Java 8 users will always see this message.

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
turkeylurkey: There are about 6 different exceptions including `NoSuchMethodException` we would need to handle if we don't catch all exceptions. Do we want to handle each separately? I'm not sure what this adds. 
Do people normally run with -i? Also, if we want to see the debug output how do we make -d work?

ericglau: If we catch all exceptions, then even all runtime exceptions (including null pointers) would be suppressed, so there may be a point where nothing is actually working but the only thing the user sees is the above message.

To keep the implementation simple, you could catch `NoSuchMethodException` in which case it should print the above message, and separately catch all other exceptions in which case at least log an error (if those exceptions are not expected)?

ericglau: Maven default log visibility is INFO, Gradle default is LIFECYCLE (INFO is hidden).  Regardless, here we should log at whatever level makes sense.  INFO if it provides benefit to the user, DEBUG otherwise.  To enable debug visibility, -X works for Maven, -d should work for Gradle but the plugin seems to suppress it (that may be a separate issue).

turkeylurkey: Could you clarify the treatment of the expected (on Java 8) exception vs. the unexpected exceptions?

ericglau: e.g. something like
```
} catch (NoSuchMethodException e) { // ignore if no method found in Java 8.
            debug(e.getClass().getName() + " trying to invoke java.net.URLConnection.setDefaultUseCaches(S,b) in disable");
} catch (Exception e) {  // warn if some other exception occurred
            warn("Could not disable caching for URLConnection: " + e.getMessage());
            debug(e.getClass().getName() + " trying to invoke java.net.URLConnection.setDefaultUseCaches(S,b) in disable", e);
}
```

PR #132, opened 2020-03-12
702 properties changes

TrevCraw:
https://github.com/OpenLiberty/ci.maven/issues/702

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1066,8 +1066,6 @@ public abstract class DevUtil {
     File bootstrapPropertiesFileParent;
     File jvmOptionsFile;
     File jvmOptionsFileParent;
-    File serverEnvFile;
-    File serverEnvFileParent;
     File buildFile;
     List<String> artifactPaths;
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1066,22 +1066,19 @@ private void readInput() {
...
-    File serverEnvFileParent;
     File buildFile;
     List<String> artifactPaths;
 
     // The serverXmlFile parameter can be null when using the server.xml from the
     // configDirectory, which has a default value.
     public void watchFiles(File buildFile, File outputDirectory, File testOutputDirectory,
             final ThreadPoolExecutor executor, List<String> artifactPaths, File serverXmlFile,
-            File bootstrapPropertiesFile, File jvmOptionsFile, File serverEnvFile) throws Exception {
+            File bootstrapPropertiesFile, File jvmOptionsFile) throws Exception {
```
cherylking: Seeing all of these changes, I'm beginning to wonder if passing in a File[] or Set<File> would be better than passing in a bunch of specifically named File parameters. Thoughts?

ericglau: bootstrapPropertiesFile and jvmOptionsFile can be treated the same way, but serverXmlFile and serverEnvFile need to be handled differently.  In this case, using an array or set of 2 files (for bootstrapPropertiesFile and jvmOptionsFile) doesn't provide much benefit over the current way.

PR #133, opened 2020-04-14
Auto detect file tracking method

TrevCraw:
Fixes https://github.com/OpenLiberty/ci.maven/issues/748

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -241,6 +241,10 @@ public abstract class DevUtil {
      */
     public abstract String getServerStartTimeoutExample();
 
+    private enum FileTrackMode {
+        NOT_SET, FILE_WATCHER, POLLING
+    }
+
     private File serverDirectory;
     private File sourceDirectory;
     private File testSourceDirectory;
@@ -274,6 +278,7 @@ public abstract class DevUtil {
     private boolean gradle;
     private boolean polling;
     private long pollingInterval;
+    private FileTrackMode trackingMode;
 
     public DevUtil(File serverDirectory, File sourceDirectory, File testSourceDirectory, File configDirectory,
             List<File> resourceDirs, boolean hotTests, boolean skipTests, boolean skipUTs, boolean skipITs,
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1288,6 +1304,16 @@ public void watchFiles(File buildFile, File outputDirectory, File testOutputDire
                     Thread.sleep(pollingInterval);
                 }
             }
+        } finally {
+            if (watcher != null) {
+                try {
+                    watcher.close();
+                } catch (IOException e) {
+                    // TODO Auto-generated catch block
+                    e.printStackTrace();
```
ericglau, line 1313: Log the exception using `error(.....)`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1351,50 +1378,74 @@ public boolean accept(File file) {
...
+                            if (watcher != null) {
+                                watcher.close();
+                            }
+                        } else {
+                            trackingMode = FileTrackMode.FILE_WATCHER;
+                            disablePolling();
+                        }
+                    } catch (Exception e) {
+                        // TODO Auto-generated catch block
+                        e.printStackTrace();
```
ericglau, line 1444: Log the exception using `error(.....)`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: How is this?
```
catch (IOException e) {
                    error("An error occurred attempting to close the file watcher. " + e.getMessage());
                }
```

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: How is this?
```
catch (Exception e) {
                        error("An error occured attempting to retrieve the watch key or close the file watcher. " + e.getMessage());
                    }
```

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: Also include the throwable as the second parameter so the stacktrace can be logged. e.g. use `error(String msg, Throwable e)`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: Also include the throwable as the second parameter so the stacktrace can be logged. e.g. use `error(String msg, Throwable e)`

PR #142, opened 2020-05-29
Generate loose application with filenames containing a variable

turkeylurkey:
Specify variable name in .war.xml file
Pass the project directory to the DevUtil object for use launching a container.


src/main/java/io/openliberty/tools/common/plugins/config/LooseConfigData.java
```diff
@@ -25,13 +25,18 @@
 import org.w3c.dom.Element;
 
 public class LooseConfigData extends XmlDocument {
-    
+
     private String projectRoot = null;
+    private String sourceOnDiskName = "/devmode";
```
ericglau, line 30: I don't think the default value should be "/devmode" here, since this class is not specific to dev mode.  Supposedly we expect that `projectRoot` and `sourceOnDiskName` are either both set, or both not set.  Either somehow enforce that rule or change line 113 to also check if `sourceOnDiskName` is set, otherwise fallback to the canonical path using line 116.

turkeylurkey: I think it is clearest if we initialize both fields to null and check on line 113 to make sure both have been set to definitive values. I'll also comment both setters to make this clear.

src/main/java/io/openliberty/tools/common/plugins/config/LooseConfigData.java
```diff
@@ -27,12 +27,14 @@
 public class LooseConfigData extends XmlDocument {
 
     private String projectRoot = null;
-    private String sourceOnDiskName = "/devmode";
+    private String sourceOnDiskName = null;
 
+    // Set both projectRoot and sourceOnDiskName to control the name used when an element is added.
```
ericglau, line 32: Use Javadoc style comments e.g.
```
/**
 *
 */
```

src/main/java/io/openliberty/tools/common/plugins/config/LooseConfigData.java
```diff
@@ -27,12 +27,14 @@
...
     public void setProjectRoot(String root) {
         projectRoot = root;
     }
 
+    // Set both projectRoot and sourceOnDiskName to control the name used when an element is added.
```
ericglau, line 37: Use Javadoc style comments e.g.
```
/**
 *
 */
```

turkeylurkey: Ok, also updated copyright year.

PR #145, opened 2020-06-11
Incorporate feedback from the design review

turkeylurkey:
Use a more descriptive variable name.
Define the variable in a server xml file and override it in the container with a Liberty command line argument.


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -777,7 +777,7 @@ private String getContainerCommand() {
         // Server 'run' command
         command.append(" /opt/ol/wlp/bin/server run defaultServer");
         // All the Liberty options:
-        command.append(" -D"+DEVMODE_PROJECT_ROOT+"="+DEVMODE_DIR_NAME);
+        command.append(" -- --"+DEVMODE_PROJECT_ROOT+"="+DEVMODE_DIR_NAME);
```
ericglau, line 780: Why does this have "-- --"

turkeylurkey: The Liberty syntax is literally '--' to end one set of options and enter a different parsing mode then '--variablename=value' for the actual definition(s).

ericglau: It would be good to add this as a comment in the code for clarity.

turkeylurkey: By way of explanation: when you use -Dkey=val that is passed to the JVM as a system property and probably needs to appear before the --. The vars after the -- are special vars that can override server.xml. The Java system properties have less priority than server.xml.

PR #149, opened 2020-06-16
Stop and start the container when a file is changed.

turkeylurkey:
For the issue https://github.com/OpenLiberty/ci.maven/issues/800

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -827,8 +827,10 @@ public abstract class DevUtil {
             }
         }
         libertyCreate();
-        // On containerized runtime local dev mode can't install feature but install in local server for consistency
-        libertyInstallFeature();
+        if (!container) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -796,47 +800,39 @@ private String getContainerCommand() {
...
+                    }
                 }
             }
         }
         libertyCreate();
-        if (!container) {
-            // local dev mode can't install feature on containerized runtime, unless we exec into the container
-            libertyInstallFeature();
-        }
+        // On containerized runtime local dev mode can't install feature but install in local server for consistency
```
ericglau, line 830: For container mode, I think we should still skip installing features for now.  There's no need to install locally for consistency, since if dev mode is started locally afterwards then the local features will be installed at that time.

turkeylurkey: Agreed.

PR #150, opened 2020-06-18
Run container in foreground, print logs to console and handle error messages

ericglau:
Fixes https://github.com/OpenLiberty/ci.maven/issues/820

Remaining issues: 
- ~~Gradle does not include the output in its console~~ Fixed
- ~~When exiting dev mode, it sometimes prints `[INFO] Error running docker command, return value=141`~~ Fixed
- ~~Handle thread safety for dockerRunProcess object~~ I don't think this is needed since it's already marked volatile so it should be consistent between different threads.  startContainer() just starts the process in a separate thread, and stopContainer() just stops the process if it is running.
- ~~If the process's output is not redirected by `processBuilder.redirectOutput(Redirect.INHERIT);`, somehow stopping the container is not synchronous which causes errors during container restart.~~ Fixed


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -673,26 +708,53 @@ private void buildDockerImage() throws PluginExecutionException {
...
-            error("Error starting container: " + r.getMessage());
-            throw r;
+            String startContainerCommand = getContainerCommand();
+            debug("startContainer, cmd="+startContainerCommand);
+
+            ProcessBuilder processBuilder = new ProcessBuilder();
+            processBuilder.command(getCommandTokens(startContainerCommand));
+            processBuilder.redirectOutput(Redirect.INHERIT);
+            dockerRunProcess = processBuilder.start();
+            dockerRunProcess.waitFor();
```
turkeylurkey, line 734: Is the expected use case that if the container fails to start the user hits ctrl-c and investigates? Otherwise I would have expected a 5 minute timeout and an error message if the container didn't start.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -673,26 +708,53 @@ private void buildDockerImage() throws PluginExecutionException {
...
     private void stopContainer() {
         try {
-            info("stopping docker container: " + containerID);
-            if (containerID != null) {
-                String s = execDockerCmd("docker stop " + containerID, 30);
+            info("Stopping docker container...");
+            if (dockerRunProcess != null) {
+                // Emulate docker stop with SIGTERM, wait up to 10 seconds, then SIGKILL
+                dockerRunProcess.destroy();
+                dockerRunProcess.waitFor(10, TimeUnit.SECONDS);
```
turkeylurkey, line 766: If I kill the container it sometimes takes about 30s to disappear. Could you make this timeout longer?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: If the container fails to start, the `waitFor()` should return and it should enter the following `if (dockerRunProcess.exitValue() != 0` section.  

Or if you mean when the server takes a long time to start, serverStart()'s `String startMessage = serverTask.waitForStringInLog(START_SERVER_MESSAGE_PREFIX, serverStartTimeoutMillis, messagesLogFile);` section should check for the message in messages.log and timeout.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: `docker stop`'s default timeout is 10 seconds though. https://docs.docker.com/engine/reference/commandline/stop/

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: Ok, I thought we were on the dev mode "UI" thread at this point and we were going to hang. This looks good.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: Ok, looks good.

ericglau: I realized that waitFor() returned immediately once I changed the implementation in the container thread to use a reader to log its content to Maven/Gradle instead of setting `processBuilder.redirectOutput(Redirect.INHERIT);`.  This caused server restart to fail since that meant `stopContainer()` was no longer synchronous.  So I changed it now to use `docker ps` and then `docker stop`, which is synchronous.  Can you please take another look?

PR #153, opened 2020-06-23
Remove WAR file lines and mount .war.xml

TrevCraw:
Fixes https://github.com/OpenLiberty/ci.maven/issues/819


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1075,6 +1131,20 @@ public void cleanUpTempConfig() {
...
 
+    public void cleanUpTempDockerfile() {
+        if (tempDockerfilePath != null) {
+            File tempDockerfile = tempDockerfilePath.toFile();
+            if (tempDockerfile.exists()) {
+                try {
+                    Files.delete(tempDockerfilePath);
+                    debug("Sucessfully deleted dev mode temporary Dockerfile");
+                } catch (IOException e) {
+                    error("Could not delete dev mode temporary Dockerfile");
```
ericglau, line 1142: Include the exception message so there is some indication on why it couldn't delete.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -518,7 +520,12 @@ public void startServer() throws PluginExecutionException {
             File dockerfileToUse = dockerfile != null ? dockerfile : defaultDockerfile;
             debug("Dockerfile to use: " + dockerfileToUse);
             if (dockerfileToUse.exists()) {
-                buildDockerImage();
+                List<String> dockerfileLines;
+                File tempDockerfile;
+                dockerfileLines = readDockerfile(dockerfileToUse);
+                dockerfileLines = removeWarFileLines(dockerfileLines);
+                tempDockerfile = createTempDockerfile(dockerfileLines);
```
ericglau, line 527: These two variables could be declared on the same lines as their assignment.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -686,24 +693,72 @@ public void onFileChange(File file) {
...
+        List<String> warFileLines = new ArrayList<String>();
+        for (String line : dockerfileLines) {
+            // Remove white space from the beginning and end of the line
+            String trimLine = line.trim();
+            if (!trimLine.startsWith("#") && trimLine.contains(".war")) {
+                // Break the Dockerfile line down into segments based on any amount of whitespace.
+                // The command must be to the left of any comments.
+                String[] cmdSegments = trimLine.split("#")[0].split("\\s+");
+                // if the line starts with COPY and the second to last segment ends with ".war", it is a WAR file COPY line
+                if (cmdSegments[0].equals("COPY") && cmdSegments[cmdSegments.length - 2].endsWith(".war")) {
```
ericglau, line 719: This fine for now, but it may have a problem if the destination path has whitespaces in it.  We can handle that as a separate issue.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -686,24 +693,72 @@ public void onFileChange(File file) {
...
+            tempDockerfile = File.createTempFile("tempDockerfile", "");
+            debug("temp Dockerfile: " + tempDockerfile);
+            tempDockerfilePath = tempDockerfile.toPath();
+            // set the tempDockerfile to be deleted when the JVM exits
+            tempDockerfile.deleteOnExit();
+            Files.write(tempDockerfile.toPath(), dockerfileLines, StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            error("Failed to create temp Dockerfile");
+            //TODO: what direction should we give the user here?
+            throw new PluginExecutionException("Could not create temp Dockerfile: ", e);
```
ericglau, line 743: Also append e.getMessage() to "Could not create temp Dockerfile: "

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: I'm not sure why it wouldn't be able to be deleted, other than it not existing (which we check for). I just mirrored this function based off of `cleanUpTempConfig()`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: Good point. Looks like quotes are required in that case, so maybe we can use that to help with the regex. https://docs.docker.com/engine/reference/builder/#copy

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: It's fine as is, but having the error included would be better for diagnosis (e.g. maybe if the file was somehow still in use).

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: Opened https://github.com/OpenLiberty/ci.maven/issues/856 to track the issue

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: I suggest using e.getMessage() in case it has an explanation.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey, line 715: Docker allows the command to be lower case: "The instruction is not case-sensitive. However, convention is for them to be UPPERCASE to distinguish them from arguments more easily." Also, Windows allows the .WAR file name to be upper case so we should use equalsIgnoreCase().

PR #154, opened 2020-06-24
Mount files from COPY lines

TrevCraw:
Fixes https://github.com/OpenLiberty/ci.maven/issues/845

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -738,9 +738,8 @@ public abstract class DevUtil {
                 // If the line starts with COPY
                 if (cmdSegments[0].equalsIgnoreCase("COPY")) {
                     if (cmdSegments.length < 3) {
-                        error("A COPY line in the Dockerfile has incorrect syntax: '" + line + "'. There must be at least " + 
-                        "two arguments, a source path and a destination path.");
-                        throw new PluginExecutionException("Cannot build docker image with incorrect COPY line syntax.");
+                        throw new PluginExecutionException("Incorrect syntax on this line in the Dockerfile: '" + line + 
+                        "'. There must be at least two arguments for the COPY command, a source path and a destination path.");
                     }
                     String src = cmdSegments[cmdSegments.length - 2];
                     String dest = cmdSegments[cmdSegments.length - 1];
@@ -764,8 +763,7 @@ public abstract class DevUtil {
             return false;
         }
         else if (!srcMountFile.exists()) {
-            error(srcMountFile + " does not exist and will not be copied to the image.");
-            throw new PluginExecutionException("Cannot build docker image with missing file " + srcMountFile);
+            throw new PluginExecutionException("Cannot build docker image with missing file: " + srcMountFile);
         }
         return true;
     }
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -296,6 +296,8 @@ public abstract class DevUtil {
     private String dockerRunOpts;
     private volatile Process dockerRunProcess;
     private File defaultDockerfile;
+    private List<String> srcMount;
+    private List<String> destMount;
 
     public DevUtil(File serverDirectory, File sourceDirectory, File testSourceDirectory, File configDirectory, File projectDirectory,
             List<File> resourceDirs, boolean hotTests, boolean skipTests, boolean skipUTs, boolean skipITs,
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -722,6 +726,57 @@ public void onFileChange(File file) {
...
+            // Remove white space from the beginning and end of the line
+            String trimLine = line.trim();
+            if (!trimLine.startsWith("#")) {
+                // Break the Dockerfile line down into segments based on any amount of whitespace.
+                // The command must be to the left of any comments.
+                String[] cmdSegments = trimLine.split("#")[0].split("\\s+");
+                // If the line starts with COPY
+                if (cmdSegments[0].equalsIgnoreCase("COPY")) {
+                    String src = cmdSegments[cmdSegments.length - 2];
+                    String dest = cmdSegments[cmdSegments.length - 1];
```
ericglau, line 741: Validate the length of cmdSegments, otherwise this may hit an array index out of bounds exception.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -722,6 +726,57 @@ public void onFileChange(File file) {
...
+        }
+        return true;
+    }
+
+    private String formatDestMount(String destMountString, String srcMountString) {
+        // Cannot mount a file onto a directory, so must add a filename to the end of the destination argument for mounting
+        if (destMountString.endsWith("/") || destMountString.endsWith("\\")) {
+            // Split on forward and backward slashes
+            String [] segments = srcMountString.split("/|\\\\");
+            destMountString += segments[segments.length -1];
```
ericglau, line 774: Would it be simpler to get a File or Path object from the source mount, and get the filename from that object?

Although the COPY source might contain wildcards according to  https://docs.docker.com/engine/reference/builder/#copy (should have that as a separate issue)

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: And if the length is not >=3? Do we print an error and throw an Exception?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: I can try that method out and see how it goes.

As for the wildcards, I agree we need to handle that. We also have a few other scenarios to consider as well for both WAR file and COPY lines.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: Yes, since COPY requires src and dest

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -737,6 +737,11 @@ public void onFileChange(File file) {
                 String[] cmdSegments = trimLine.split("#")[0].split("\\s+");
                 // If the line starts with COPY
                 if (cmdSegments[0].equalsIgnoreCase("COPY")) {
+                    if (cmdSegments.length < 3) {
+                        error("A COPY line in the Dockerfile has incorrect syntax: '" + line + "'. There must be at least " + 
+                        "two arguments, a source path and a destination path.");
+                        throw new PluginExecutionException("Cannot build docker image with incorrect COPY line syntax.");
```
ericglau, line 743: Include the above error message in the exception message or its cause.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -296,6 +296,8 @@ public abstract boolean recompileBuildFile(File buildFile, List<String> artifact
     private String dockerRunOpts;
     private volatile Process dockerRunProcess;
     private File defaultDockerfile;
+    private List<String> srcMount;
+    private List<String> destMount;
```
turkeylurkey, line 300: Since src and dest are two things that always go together I suggest these arrays be replaced by one HashMap. This would avoid the values getting out of synch if something unforeseen caused one of the values in one of the arrays to be dropped or not inserted.

TrevCraw: Originally, I was actually thinking of using a HashMap. But the issue with using a HashMap is technically a `src` could be repeated, because a user could potentially copy the same file to multiple locations. That would lead to duplicate keys, which you cannot have in a regular <String, String> HashMap. So I looked into how to handle multiple keys in a HashMap, which led to things I thought were maybe too excessive for this particular use case.

We can discuss further if you would like. @ericglau  any thought on this?

ericglau: HashMap would not work for all cases here, such as the below.
```
COPY a b
COPY a c
```
A better data structure could be used though to prevent them getting out of sync, such as a 2d array or an array of objects (where each object has src and dest).  But I think it's ok as is.

PR #155, opened 2020-06-24
Tech preview for Dev Mode with Container support

ericglau:
Part of https://github.com/OpenLiberty/ci.maven/issues/659


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -465,6 +496,24 @@ public int countApplicationUpdatedMessages() {
...
+     * If the server directory path cannot be resolved, get log file from the server task instead.
+     * 
+     * @param serverTask the server task
+     * @return the messages log file for the server
+     */
+    private File getMessagesLogFile(ServerTask serverTask) {
+        File logFile;
+        try {
+            String logsDirectory = serverDirectory.getCanonicalPath() + "/logs";
+            logFile = new File(logsDirectory, "messages.log");
```
cherylking, line 510: I think you can do this instead and avoid calling getCanonicalPath(): `logFile = new File(serverDirectory, "/logs/messages.log");`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -465,6 +496,24 @@ public int countApplicationUpdatedMessages() {
...
+        } catch (IOException e) {
+            logFile = serverTask.getLogFile();
```
cherylking, line 512: guard against serverTask == null

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -473,6 +522,18 @@ public int countApplicationUpdatedMessages() {
      *                                  failed.
      */
     public void startServer() throws PluginExecutionException {
+        if (container) {
+            File dockerfileToUse = dockerfile != null ? dockerfile : defaultDockerfile;
+            debug("Dockerfile to use: " + dockerfileToUse);
+            if (dockerfileToUse.exists()) {
+                List<String> dockerfileLines = removeCopyLines(removeWarFileLines(readDockerfile(dockerfileToUse)), dockerfileToUse.getParent());
```
cherylking, line 529: In general, nesting method calls like this makes support difficult. Why not let the createTempDockerfile method call each of those sequentially and return the temp dockerfile. Maybe call it prepareTempDockerfile to indicate it does some processing before creating it.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -566,39 +644,366 @@ public void onFileChange(File file) {
...
+            // set environment variables in the container to ensure debug mode does not suspend the server, and to enable a custom debug port to be used
+            command.append(" -e WLP_DEBUG_SUSPEND=n -e WLP_DEBUG_ADDRESS=" + libertyDebugPort);
+        }
+
+        // mount .war.xml potential directories - override /config/apps and /config/dropins
+        command.append(" -v " + serverDirectory + "/apps:/config/apps");
+        command.append(" -v " + serverDirectory + "/dropins:/config/dropins");
+
+        // mount the loose application resources in the container
+        command.append(" -v "+projectDirectory.getAbsolutePath()+":"+DEVMODE_DIR_NAME);
```
cherylking, line 978: getCanonicalPath()

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -566,39 +644,366 @@ public void onFileChange(File file) {
...
+
+        // mount the server logs directory over the /logs used by the open liberty container as defined by the LOG_DIR env. var.
+        command.append(" -v "+serverDirectory.getAbsolutePath()+"/logs:/logs");
```
cherylking, line 981: getCanonicalPath()

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -763,9 +1179,23 @@ public void cleanUpTempConfig() {
             if (tempConfig.exists()) {
                 try {
                     FileUtils.deleteDirectory(tempConfig);
-                    debug("Sucessfully deleted liberty:dev temporary configuration folder");
+                    debug("Successfully deleted liberty:dev temporary configuration folder");
                 } catch (IOException e) {
-                    error("Could not delete liberty:dev temporary configuration folder");
+                    error("Could not delete liberty:dev temporary configuration folder: " + e.getMessage());
```
cherylking, line 1184: warning instead of error

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -763,9 +1179,23 @@ public void cleanUpTempConfig() {
...
+
+    public void cleanUpTempDockerfile() {
+        if (tempDockerfilePath != null) {
+            File tempDockerfile = tempDockerfilePath.toFile();
+            if (tempDockerfile.exists()) {
+                try {
+                    Files.delete(tempDockerfilePath);
+                    debug("Successfully deleted dev mode temporary Dockerfile");
+                } catch (IOException e) {
+                    error("Could not delete dev mode temporary Dockerfile: " + e.getMessage());
```
cherylking, line 1198: warning instead of error

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
cherylking: When using the defaultDockerfile, I think calling removeWarFileLines and removeCopyLines is unnecessary isn't it? And how exactly would the defaultDockerfile not exist?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -566,39 +644,366 @@ public void onFileChange(File file) {
...
+        List<String> warFileLines = new ArrayList<String>();
+        for (String line : dockerfileLines) {
+            // Remove white space from the beginning and end of the line
+            String trimLine = line.trim();
+            if (!trimLine.startsWith("#") && trimLine.toLowerCase().contains(".war")) {
+                // Break the Dockerfile line down into segments based on any amount of whitespace.
+                // The command must be to the left of any comments.
+                String[] cmdSegments = trimLine.split("#")[0].split("\\s+");
+                // if the line starts with COPY and the second to last segment ends with ".war", it is a WAR file COPY line
+                if (cmdSegments[0].equalsIgnoreCase("COPY") && cmdSegments[cmdSegments.length - 2].toLowerCase().endsWith(".war")) {
```
cherylking, line 721: What if the line does not have 3 segments? The `cmdSegments[cmdSegments.length-2]` part of this will get an ArrayIndexOutOfBoundsException.

ericglau: This is checked in the part where it has:
```
                    if (cmdSegments.length < 3) {
                        throw new PluginExecutionException("Incorrect syntax on this line in the Dockerfile: '" + line + 
                        "'. There must be at least two arguments for the COPY command, a source path and a destination path.");
                    }
```

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: This creates the command to pass to Docker, and it is not used for path comparison.  So absolute path should work and it just leaves the path resolution to the command line or Docker.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: This creates the command to pass to Docker, and it is not used for path comparison.  So absolute path should work and it just leaves the path resolution to the command line or Docker.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: defaultDockerfile is not a default dockerfile that we create.  Instead, it just represents the default dockerfile location that we are expecting.  It doesn't exist if the user does not have a Dockerfile in their project.  So regardless, we still need to call  removeWarFileLines and removeCopyLines since the Dockerfile is a user-provided file.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
cherylking: That check is in another method and does not apply to here. This line will fail with the exception before it ever gets to that other method.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corporation 2019.
+ * (C) Copyright IBM Corporation 2020.
```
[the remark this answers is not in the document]
ericglau: done

src/main/java/io/openliberty/tools/common/plugins/config/LooseConfigData.java
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corporation 2017.
+ * (C) Copyright IBM Corporation 2017-2020.
```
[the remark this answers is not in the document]
ericglau: done

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -314,6 +336,14 @@ public DevUtil(File serverDirectory, File sourceDirectory, File testSourceDirect
         } else {
             this.trackingMode = FileTrackMode.NOT_SET;
         }
+        this.container = container;
+        this.dockerfile = dockerfile;
+        this.dockerRunOpts = dockerRunOpts;
+        if (projectDirectory != null) {
+            this.defaultDockerfile = new File (projectDirectory.getAbsolutePath() + "/Dockerfile");
```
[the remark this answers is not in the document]
ericglau: Changed to `new File(projectDirectory, "Dockerfile");`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -314,6 +336,14 @@ public DevUtil(File serverDirectory, File sourceDirectory, File testSourceDirect
...
+        }
+        this.srcMount = new ArrayList<String>();
```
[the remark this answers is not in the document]
ericglau: done

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: done

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: done

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: done

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: Changed to prepareTempDockerfile and sequential calls from there

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: done

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: Added validation to this method

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -778,8 +779,13 @@ private String formatDestMount(String destMountString, File srcMountFile) {
         return destMountString;
     }
 
-    private File createTempDockerfile(List<String> dockerfileLines) throws PluginExecutionException {
+    private File prepareTempDockerfile(File dockerfile) throws PluginExecutionException {
         // Create a temp Dockerfile to build image from
+
+        List<String> dockerfileLines = readDockerfile(dockerfile);
+        dockerfileLines = removeWarFileLines(dockerfileLines);
```
cherylking, line 786: This is a nit, but there is no need to pass back the List from these methods (removeWarFileLines and removeCopyLines). It was passed in and modified.

ericglau:
Added a message pointing to docs with an example Dockerfile in case there is no Dockerfile found.

PR #157, opened 2020-06-26
Handle container checks

ericglau:
- libertyInstallFeature() is abstract, so check for container option and skip installing features (for now) in Maven/Gradle instead of here.
- getMessagesLogFile() should avoid using serverTask to get the log file in container mode, the ServerTask did not have its initTask() called yet.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -502,7 +502,7 @@ public abstract class DevUtil {
      */
     private File getMessagesLogFile(ServerTask serverTask) {
         File logFile;
-        if (serverTask != null) {
+        if (!container && serverTask != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -502,7 +502,7 @@ public int countApplicationUpdatedMessages() {
      */
     private File getMessagesLogFile(ServerTask serverTask) {
         File logFile;
-        if (serverTask != null) {
+        if (!container && serverTask != null) {
```
cherylking, line 505: So I think if container is true, we always want to not use the serverTask. And if container is not true, we always want to use serverTask. Right? I would rewrite as follows:
```
if (container) {
   logFile = new File(serverDirectory, "logs/messages.log");
} else {
   logFile = serverTask.getLogFile();
}
```

PR #158, opened 2020-06-27
Handle Dockerfiles with multiline and comments

ericglau:
Handle Dockerfile with COPY commands that span multiple lines and have comments in between

Fixes https://github.com/OpenLiberty/ci.maven/issues/863

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -811,8 +811,7 @@ public abstract class DevUtil {
         dockerfileLines.removeAll(warFileLines);
     }
 
-    private void removeCopyLines(List<String> dockerfileLines, String buildContext) throws PluginExecutionException {
-        List<String> copyLines = new ArrayList<String>();
+    private void processCopyLines(List<String> dockerfileLines, String buildContext) throws PluginExecutionException {
         for (String line : dockerfileLines) {
             // Remove white space from the beginning and end of the line
             String trimLine = line.trim();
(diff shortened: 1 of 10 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -703,6 +703,50 @@ public void onFileChange(File file) {
...
+     * Combine multi-line commands into single lines
+     */
+    protected static List<String> getCombinedLines(List<String> dockerfileLines) throws PluginExecutionException {
+        List<String> result = new ArrayList<String>();
+        int i = 0;
+        while (i < dockerfileLines.size()) {
+            String pendingLine = dockerfileLines.get(i).trim();
+            int multilineIndex;
+            int j = i+1;
+            while ((multilineIndex = pendingLine.indexOf("\\")) >= 0 && j < dockerfileLines.size()) {
```
turkeylurkey, line 737: Technically you can use \ in a filename on Linux. Can we document this method only handles "clean" lines (no comments or trailing whitespace) so that we can check for lines that end with '\\'?

turkeylurkey:
Perhaps we should make a tactical decision regarding the tech preview to support COPY statements like we see in the guide and for the full release we support a more robust implementation. We should also handle the escape character in the preview because it is important for Windows.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java

ericglau:
Added support for line breaks and the `escape` directive.  For now we cannot handle environment variables.  After the preview we can look at better handling for Dockerfile parsing.

ericglau:
Added handling multiple `<src>` to a single `<dest>`.  Added checks to give warnings for wildcards or env vars.

ericglau:
The new tests failed on Appveyor, but pass when I manually run them on Windows.

Seems the second-last line is always missing when the temp Dockerfile is generated on Appveyor.

```
org.junit.ComparisonFailure: expected:<...nel-java8-openj9-ubi[
COPY --chown=1001:0  server.xml /config/
ARG VERBOSE=false]
RUN configure.sh> but was:<...nel-java8-openj9-ubi[
COPY --chown=1001:0  server.xml /config/
]
RUN configure.sh>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at io.openliberty.tools.common.plugins.util.DevUtilPrepareDockerfileTest.testPrepareDockerfile(DevUtilPrepareDockerfileTest.java:55)
	at io.openliberty.tools.common.plugins.util.DevUtilPrepareDockerfileTest.testBasicDockerfile(DevUtilPrepareDockerfileTest.java:89)
```

ericglau:
This commit https://github.com/OpenLiberty/ci.common/pull/158/commits/ff2ee9bb692f3417334437d2ba244536424bbf5f fixed the Windows Appveyor unit test failures, not sure why.

Issue #159, opened 2020-06-29
Open Liberty server does not start using devmode when messages.log format is JSON


k-rtik:
I set the following property to set `messages.log` to JSON:
`
com.ibm.ws.logging.console.format=json
`

and I get the following error when I start the server:
`
Could not parse the host name from the log message:
{"type":"liberty_message","host":"localhost","ibm_userDir":"\/Users\/kartiksaravana\/workspace\/logsinsight\/target\/liberty\/wlp\/usr\/","ibm_serverName":"defaultServer","message":"CWWKT0016I: Web application available (default_host): http:\/\/192.168.0.10:9080\/logsinsight\/","ibm_threadId":"0000003b","ibm_datetime":"2020-06-29T16:46:48.037-0400","ibm_messageId":"CWWKT0016I","module":"com.ibm.ws.http.internal.VirtualHostImpl","loglevel":"AUDIT","ibm_sequence":"1593463608037_000000000000C"}
`

I tried to make some [changes](https://github.com/k-rtik/ci.common/commit/939911fc9f3b12a7c8d3ba93415c6d6ab2dccb92#diff-075b1a05105251c31cd5c46361280926) to fix the issue:
https://github.com/k-rtik/ci.common/tree/json-message

cherylking:
Since the json format for logs is fairly new, this would be an enhancement for the Liberty Maven and Gradle Plugins to add to support this.

cherylking:
Also, I believe this would affect other parts of the plugins that look for messages in the messages.log, such as whether an application has successfully started.

cherylking:
Two things to note:

1) The property that sets json format for messages.log is actually `com.ibm.ws.logging.message.format`.
2) Setting the messages.log format to json does not cause other parts of the plugin to fail. I confirmed that the server start and application start messages are found successfully.

dshimo:
Thanks @k-rtik 
Code is merged in

PR #162, opened 2020-07-21
Implement Kartik's escape char's fix

dshimo:
Fix Issue #159


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1179,18 +1181,34 @@ private void parseHostNameAndPorts(final ServerTask serverTask, File messagesLog
...
+            protocolIndex = webAppMessage.indexOf(HTTPS_PREFIX_ESCAPED);
+            hostNameIndex = protocolIndex + HTTPS_PREFIX_ESCAPED.length();
+        }
+
+        if (protocolIndex < 0) {
+            throw new PluginExecutionException(
+                    "Could not parse the host name from the log message: " + webAppMessage);
         }
 
+        debug("Position of protocol (escaped or unescaped) in hostname: " + protocolIndex);
```
cherylking, line 1211: I don't think this debug statement is needed.

PR #163, opened 2020-07-21
Mount preview only

TrevCraw:


TrevCraw:
This code is no longer necessary since https://github.com/OpenLiberty/ci.common/pull/165 was completed.

PR #164, opened 2020-07-21
Set up permissions for the logs directory on Linux.

turkeylurkey:
https://github.com/OpenLiberty/ci.maven/issues/865

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -987,7 +987,7 @@ public abstract class DevUtil {
         Process p = Runtime.getRuntime().exec(cmd);
         p.waitFor(5, TimeUnit.SECONDS);
         if (p.exitValue() != 0) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -970,6 +983,14 @@ public void run() {
         }
     }
 
+    private void runCMD(String cmd) throws IOException, InterruptedException {
+        Process p = Runtime.getRuntime().exec(cmd);
+        p.waitFor(5, TimeUnit.SECONDS);
+        if (p.exitValue() != 0) {
+            debug("Error running command:" + cmd + ", return value=" + p.exitValue());
```
ericglau, line 990: Should this log as error?  Or do we expect that some of the commands will fail?

turkeylurkey: I agree, the user may need to adjust the directories.

PR #167, opened 2020-07-29
Set up file permissions on Linux to start a container as a non-root user

turkeylurkey:
Sometimes when you are a non-root user you have permission to modify a file but sometimes the file was created by the docker container. Thus you can get errors trying to modify a file the second time you run dev mode. Thus some error reporting is removed and only available in debug mode. The alternative would be to loop through all the files and not touch files that could not be changed just for the purpose of not generating an error message. 

Fixes https://github.com/OpenLiberty/ci.maven/issues/888

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -948,12 +948,11 @@ public abstract class DevUtil {
             String os = System.getProperty("os.name");
             String id = System.getProperty("user.name");
             if (os != null && os.equalsIgnoreCase("linux")) {
-                // Allow the container server to read the config files e.g. server.xml
+                // Allow the container server to read the config files like server.xml
                 runCmd(false, "chmod -R o+r " + serverDirectory);
                 // Allow the container server to read directories like apps and dropins
-                runCmd("chmod -R o+x " + serverDirectory + "/apps " +
-                    serverDirectory + "/dropins " +
-                    serverDirectory + "/configDropins");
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -947,14 +947,22 @@ private void startContainer() {
...
-            if (os != null && os.equalsIgnoreCase("linux") &&
-                id != null && id.equalsIgnoreCase("root")) {
+            if (os != null && os.equalsIgnoreCase("linux")) {
                 // Allow the container server to read the config files e.g. server.xml
-                runCMD("chmod -R o+r " + serverDirectory);
+                runCmd(false, "chmod -R o+r " + serverDirectory);
+                // Allow the container server to read directories like apps and dropins
+                runCmd("chmod -R o+x " + serverDirectory + "/apps " +
+                    serverDirectory + "/dropins " +
+                    serverDirectory + "/configDropins");
```
ericglau, line 956: Is the intent to set the o+r on all of the _files_ under serverDirectory, but set o+x on all of its sub-directories only (excluding their files)?  Because the current code would also set o+x on the _files_ inside `apps`, `dropins` etc. which seems excessive.  Should it limit the o+x to only directories e.g. by using `find -type d` as shown [here](https://unix.stackexchange.com/a/431889)?

turkeylurkey: Good point, I wanted to avoid setting x on files for security reasons. I'll use find and see if I can manage the logs directory as well.

PR #168, opened 2020-07-30
Copy container error stream to devc output

ericglau:
Addresses https://github.com/OpenLiberty/ci.maven/issues/889 and https://github.com/OpenLiberty/ci.gradle/issues/505


turkeylurkey:
The 'firstLine' treatment is pretty cool.

PR #171, opened 2020-08-12
Detect when a config file is edited with vim on Linux and

turkeylurkey:
restart the container when using containers.

This code assumes all the files in the config directory in the container are also in the Dockerfile. 
restartServer() defaults to *no* docker build while startServer() defaults to doing a docker build.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2323,7 +2323,7 @@ public abstract class DevUtil {
                     enableServerDebug(false);
                 }
                 if (container && System.getProperty("os.name").equalsIgnoreCase("linux")) {
-                    info("Now restarting the container for this change to take effect which is required on Linux.");
+                    info("Restarting the container for this change to take effect.");
                     // Allow a 1 second grace period to replace the file in case the user changes the file with a script or a tool like vim.
                     try {
                         Thread.sleep(1000);
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2323,12 +2323,12 @@ public abstract class DevUtil {
                     enableServerDebug(false);
                 }
                 if (container && System.getProperty("os.name").equalsIgnoreCase("linux")) {
-                    info("Restarting the container is required on Linux for it to take effect.");
+                    info("Now restarting the container for this change to take effect which is required on Linux.");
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2308,6 +2322,16 @@ private void processFileChanges(
                     // re-enable debug variables in server.env
                     enableServerDebug(false);
                 }
+                if (container && System.getProperty("os.name").equalsIgnoreCase("linux")) {
+                    info("Restarting the container is required on Linux for it to take effect.");
```
ericglau, line 2326: This should be rephrased, since it currently sounds as if it's asking the user to restart the container, when in fact it is doing it automatically.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2308,6 +2322,16 @@ private void processFileChanges(
...
+                    // Allow a 1 second grace period to replace the file in case the user changes the file with a script or a tool like vim.
+                    try {
+                        Thread.sleep(1000);
+                    } catch (InterruptedException e) {
+                        debug("Unexpected InterruptedException handling config file deletion.");
```
ericglau, line 2331: Use `debug("Unexpected InterruptedException handling config file deletion.", e);`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: 👍

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: 👍

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2323,12 +2323,12 @@ private void processFileChanges(
                     enableServerDebug(false);
                 }
                 if (container && System.getProperty("os.name").equalsIgnoreCase("linux")) {
-                    info("Restarting the container is required on Linux for it to take effect.");
+                    info("Now restarting the container for this change to take effect which is required on Linux.");
```
ericglau, line 2326: The message still seems a bit verbose.  Does the user care on what OS the restart is required?  If they are on Linux then they should not be concerned about other OSes.  Perhaps we should just say "Restarting the container for this change to take effect..."

turkeylurkey: Ok.

PR #172, opened 2020-08-13
Show docker build command and better layout for the docker run command.

turkeylurkey:
Fixes https://github.com/OpenLiberty/ci.maven/issues/901

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -971,7 +971,7 @@ public abstract class DevUtil {
                 }
             }
 
-            info("Starting Docker image...");
+            info("Starting Docker container...");
             String startContainerCommand = getContainerCommand();
             info(startContainerCommand);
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -970,8 +971,9 @@ private void startContainer() {
                 }
             }
 
+            info("Starting Docker image...");
             String startContainerCommand = getContainerCommand();
-            debug("startContainer, cmd="+startContainerCommand);
+            info(startContainerCommand);
```
ericglau, line 976: The getContainerCommand() already prints the docker run command here: https://github.com/OpenLiberty/ci.common/blob/master/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java#L1190

Either remove the one from getContainerCommand() or remove this one to avoid printing it twice.  I'd prefer to remove the one from getContainerCommand() to be consistent with how we print the docker build command above.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -970,8 +971,9 @@ private void startContainer() {
                 }
             }
 
+            info("Starting Docker image...");
```
ericglau, line 974: Should be `"Starting Docker container..."`

turkeylurkey: Sure

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: I believe that was done already.

ericglau: Thanks, I didn't notice that.

PR #173, opened 2020-08-17
Add a parameter to control the amount of time to wait for a docker command

turkeylurkey:


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -349,6 +350,11 @@ public DevUtil(File serverDirectory, File sourceDirectory, File testSourceDirect
         if (projectDirectory != null) {
             this.defaultDockerfile = new File(projectDirectory, "Dockerfile");
         }
+        if (dockerTimeout < 1 || dockerTimeout > 3600) {
```
ericglau, line 353: It may be simpler (less confusing in the documentation) if we just don't have an upper bound.  i.e. easier to understand that it's in seconds and must be greater than zero.

TrevCraw:
@turkeylurkey  @ericglau 

Does this timeout apply to all docker commands? Does it make sense to have one timeout value for the different commands? `build, `run`, and `ps` all have different expectations on how long they may run.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: I'm just afraid they'll mistakenly put in a really big number and then docker will take a long time and then they'll complain dev mode is hanging. But that is a pathological failure so you could say an edge case. I'll remove the top limit unless you're worried about the same thing.

turkeylurkey:
Regarding the various commands I was thinking of using timeout/2 for ps and stop but then just to make it simpler I figured to use the same timeout for all of them. If dev mode "hangs" for 60 seconds I was hoping they would realize the hang was the same time as the timeout they entered.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: That is possible but I agree it's an edge case.  Even having 3600 as a limit might be misleading because users should really use a lower number and do something else to speed up the docker build (e.g. remove unnecessary things from the build context), rather than wait an hour.

turkeylurkey:
As discussed in person we'll leave the docker ps and docker stop timeouts at 10 and 30 respectively.

PR #174, opened 2020-08-17
Print improved message when docker build times out.

turkeylurkey:
Fixes https://github.com/OpenLiberty/ci.maven/issues/896

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -945,7 +945,7 @@ public abstract class DevUtil {
             debug("Docker build output: " + buildOutput);
         } catch (RuntimeException r) {
             error("Error building Docker image: " + r.getMessage());
-            throw new PluginExecutionException("Could not build Docker image using Dockerfile: " + userDockerfile.getAbsolutePath() + ". Address the following docker build error and then start dev mode again:" + r.getMessage(), r);
+            throw new PluginExecutionException("Could not build Docker image using Dockerfile: " + userDockerfile.getAbsolutePath() + ". Address the following docker build error and then start dev mode again: " + r.getMessage(), r);
         }
     }
 
@@ -1119,6 +1119,12 @@ public abstract class DevUtil {
             if (allLines.length() > 0) {
                 result = allLines.toString();
             }
+        } catch (IllegalThreadStateException  e) {
+            // the timeout was too short and the docker command has not yet completed. There is no exit value.
+            debug("IllegalThreadStateException, message="+e.getMessage());
+            error("The docker command did not complete within the timeout period: " + timeout + " seconds. " +
+                "Use the dockerTimeout option to specify a longer period or investigate issues running docker. ", e);
+            throw new RuntimeException("The docker command did not complete within the timeout period: " + timeout + " seconds. ");
         } catch (InterruptedException e) {
             // Container error, throw exception
             // If a runtime exception occurred in the server task, log and rethrow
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1119,6 +1119,12 @@ private String execDockerCmd(String command, int timeout) {
             if (allLines.length() > 0) {
                 result = allLines.toString();
             }
+        } catch (IllegalThreadStateException  e) {
+            // the timeout was too short and the docker command has not yet completed. There is no exit value.
+            debug("IllegalThreadStateException, message="+e.getMessage());
+            error("The docker command did not complete within the timeout period: " + timeout + " seconds. " +
+                "Use the dockerTimeout option to specify a longer period or investigate issues running docker. ", e);
```
ericglau, line 1126: "investigate issues running docker" sounds kind of vague.  If possible, be more specific on what the user should do.  Although this `execDockerCmd` method is used for docker build, docker ps, and docker stop so the user action may not be the same for all cases.

Or, perhaps just "investigate issues with running the docker command." if we need to be generic here.

turkeylurkey: Check out the new message.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1123,7 +1123,7 @@ private String execDockerCmd(String command, int timeout) {
             // the timeout was too short and the docker command has not yet completed. There is no exit value.
             debug("IllegalThreadStateException, message="+e.getMessage());
             error("The docker command did not complete within the timeout period: " + timeout + " seconds. " +
-                "Use the dockerTimeout option to specify a longer period or investigate issues running docker. ", e);
+                "Use the dockerTimeout option to specify a longer period or investigate performance issues with running the docker command. ", e);
```
ericglau, line 1126: Now the dockerTimeout option is not relevant to this message if the executed command was docker ps or docker stop.

turkeylurkey: Displaying a message specific to docker build....

TrevCraw:
@turkeylurkey do you have any example output for this?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1119,6 +1119,17 @@ private String execDockerCmd(String command, int timeout) {
             if (allLines.length() > 0) {
                 result = allLines.toString();
             }
+        } catch (IllegalThreadStateException  e) {
+            // the timeout was too short and the docker command has not yet completed. There is no exit value.
+            debug("IllegalThreadStateException, message="+e.getMessage());
+            if (command.startsWith("docker build")) {
+                error("The docker build command did not complete within the timeout period: " + timeout + " seconds. " +
+                    "Use the dockerBuildTimeout option to specify a longer period or " +
+                    "add files not needed in the container to the .dockerIgnore file. ", e);
```
ericglau, line 1128: Should be lowercase .dockerignore

turkeylurkey: Oops!

turkeylurkey:
Sample output:
```
[INFO] docker build -f /var/folders/p1/v7svjg7s6xngg377xtsjxcmc0000gn/T/tempDockerfile6741327853676937183 -t liberty-dev-mode /Users/paulg/WAS/devExperience/demo-devmode
[ERROR] The docker build command did not complete within the timeout period: 1 seconds. Use the dockerBuildTimeout option to specify a longer period or add files not needed in the container to the .dockerignore file.
java.lang.IllegalThreadStateException: process hasn't exited
    at java.lang.UNIXProcess.exitValue (UNIXProcess.java:421)
    at io.openliberty.tools.common.plugins.util.DevUtil.execDockerCmd (DevUtil.java:1103)
... ```

PR #177, opened 2020-08-20
Only mount the file to the container if it exists.

turkeylurkey:
On Linux if the file is missing Docker assumes you want to mount a directory on a directory so it creates the missing directory on the host and the container as required. In the case of our Dockerfile copies like server.xml or bootstrap.properties we need to check if they exist before mounting. Once we support to copy a directory of files in Dockerfile we may need to update this code. 
Fixes https://github.com/OpenLiberty/ci.gradle/issues/519

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1171,7 +1171,12 @@ public abstract class DevUtil {
 
         // mount all files from COPY commands in the Dockerfile to allow for hot deployment
         for (int i=0; i < srcMount.size(); i++) {
-            command.append(" -v " + srcMount.get(i) + ":" + destMount.get(i));
+            if (new File(srcMount.get(i)).exists()) { // only Files are in this list
+                command.append(" -v " + srcMount.get(i) + ":" + destMount.get(i));
+            } else {
+                error("A file referenced by the Dockerfile is not found: " + srcMount.get(i) +
+                    ". Update the Dockerfile or ensure the file is in the correct location.");
+            }
         }
 
         command.append(" --name " + DEVMODE_CONTAINER_NAME);


TrevCraw:
This change means an error would not occur correct? Wouldn't we want to throw an error in this case?  We should give the users an error (or a warning?) when they delete a file that is "copied" in the Dockerfile.

turkeylurkey:
When you add back the deleted file the server does not necessarily restart. 

In the use case of 'rm file' then an error message is a good idea but in the case of 'vim file' and then write file changes, vim deletes the old file and adds the new file and an error message is misleading. We should investigate how to handle the event stream.

ericglau:
In addition to Trevor's comment, should it actually fail (exit dev mode) if this error condition occurs?

TrevCraw:
@ericglau I was considering that as well. I'm not sure what the best way would be.  However, if I were a user, I might not want dev mode to exit, because maybe I just made a mistake deleting the file, or maybe I deleted the file first, and then I'm going to remove it from the Dockerfile. In both those scenarios, I'd want dev mode to keep running.

ericglau:
If you made a mistake deleting the file, and you try to add it back, I think it doesn't take effect so dev mode would need to be restarted anyways.

I agree with the second point, but if we don't exit then the user might never delete it from the Dockerfile.  Then when they start up dev mode the next time, the `docker build` command will fail because the file is missing (which is fine, but might as well fail earlier).

turkeylurkey:
True: error messages usually have recovery suggestions. I thought it too obvious but perhaps it is standard practice to add something like "Ensure all the files referenced in the Dockerfile are available on your system." Also, I'm using the term "the Dockerfile." Should we include the actual name of the Dockerfile?

TrevCraw:
@ericglau I thought the replaced mounted file not taking effect was only a Linux thing? Even if not, wouldn't a container restart be better than forcing the user to have to start dev mode again?

That's a good point for the second scenario. I think that may come down to personal preference. Not sure which way would be better.

TrevCraw:
@turkeylurkey I don't know if it's necessary, but we could say something like "your Dockerfile at /PATH/to/Dockerfile"

PR #182, opened 2020-08-26
Issue a warning if docker build takes longer than 30 secs.

turkeylurkey:
Fixes https://github.com/OpenLiberty/ci.maven/issues/903

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1115,7 +1115,7 @@ public abstract class DevUtil {
                 throw new RuntimeException(errorMessage);
             }
             if (command.startsWith("docker build")) {
-                dockerIgnoreHelpMessage(startTime);
+                checkDockerIgnore(startTime);
             }
 
             // Read all the output on stdout and return it to the caller
@@ -1154,19 +1154,24 @@ public abstract class DevUtil {
     }
 
     // Suggest a performance improvement if docker build takes too long.
-    static final long DOCKER_BUILD_SOFT_TIMEOUT = 30000;
-    private void dockerIgnoreHelpMessage(long startTime) {
+    private static final long DOCKER_BUILD_SOFT_TIMEOUT = 30000;
+    private void checkDockerIgnore(long startTime) {
         if (System.currentTimeMillis() - startTime < DOCKER_BUILD_SOFT_TIMEOUT) {
             return;
         }
         File dockerfileToUse = dockerfile != null ? dockerfile : defaultDockerfile;
         if (dockerfileToUse.exists()) {
             File dockerContext = dockerfileToUse.getParentFile();
-            debug("dockerIgnoreHelpMessage, dockerContext="+dockerContext.getAbsolutePath());
+            debug("checkDockerIgnore, dockerContext="+dockerContext.getAbsolutePath());
             File dockerIgnore = new File(dockerContext, ".dockerignore");
             if (!dockerIgnore.exists()) { // provide some advice
-                warn("The docker build command is slower than expected. You may increase performance by adding " + 
-                    "unneeded files and directories such as any Liberty runtime directories to the .dockerignore file.");
+                try {
+                    warn("The docker build command is slower than expected. You may increase performance by adding " + 
+                        "unneeded files and directories such as any Liberty runtime directories to a .dockerignore file in " +
+                        dockerContext.getCanonicalPath()+ ".");
+                } catch (IOException e) {
+                    error("Exception while retrieving the path name of the docker context directory", e);
+                }
             }
         }
     }
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1103,6 +1103,7 @@ public abstract class DevUtil {
         String result = null;
         try {
             debug("execDocker, timeout=" + timeout + ", cmd=" + command);
+            long startTime = System.currentTimeMillis();
             Process p = Runtime.getRuntime().exec(command);
             p.waitFor(timeout, TimeUnit.SECONDS);
             if (p.exitValue() != 0) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1149,6 +1153,24 @@ private String execDockerCmd(String command, int timeout) {
         return result;
     }
 
+    // Suggest a performance improvement if docker build takes too long.
+    static final long DOCKER_BUILD_SOFT_TIMEOUT = 30000;
+    private void dockerIgnoreHelpMessage(long startTime) {
```
ericglau, line 1158: Method name should be a verb that describes what it does.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1149,6 +1153,24 @@ private String execDockerCmd(String command, int timeout) {
         return result;
     }
 
+    // Suggest a performance improvement if docker build takes too long.
+    static final long DOCKER_BUILD_SOFT_TIMEOUT = 30000;
```
ericglau, line 1157: Also use `private` unless this is needed elsewhere.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1149,6 +1153,24 @@ private String execDockerCmd(String command, int timeout) {
...
+            return;
+        }
+        File dockerfileToUse = dockerfile != null ? dockerfile : defaultDockerfile;
+        if (dockerfileToUse.exists()) {
+            File dockerContext = dockerfileToUse.getParentFile();
+            debug("dockerIgnoreHelpMessage, dockerContext="+dockerContext.getAbsolutePath());
+            File dockerIgnore = new File(dockerContext, ".dockerignore");
+            if (!dockerIgnore.exists()) { // provide some advice
+                warn("The docker build command is slower than expected. You may increase performance by adding " + 
+                    "unneeded files and directories such as any Liberty runtime directories to the .dockerignore file.");
```
ericglau, line 1169: Should this mention what directory the .dockerignore file should go in? e.g. include the `dockerContext` path in the message.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: 👌

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: 👌

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: Yes, that would be helpful.

turkeylurkey:
Sample output:
```
[INFO] Building Docker image...
[INFO] docker build -f /var/folders/p1/v7svjg7s6xngg377xtsjxcmc0000gn/T/tempDockerfile6946523769638832910 -t liberty-dev-mode /Users/user1/devExperience/demo-devmode
[WARNING] The docker build command is slower than expected. You may increase performance by adding unneeded files and directories such as any Liberty runtime directories to a .dockerignore file in /Users/user1/devExperience/demo-devmode.
```

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1154,19 +1154,24 @@ private String execDockerCmd(String command, int timeout) {
...
             File dockerIgnore = new File(dockerContext, ".dockerignore");
             if (!dockerIgnore.exists()) { // provide some advice
-                warn("The docker build command is slower than expected. You may increase performance by adding " + 
-                    "unneeded files and directories such as any Liberty runtime directories to the .dockerignore file.");
+                try {
+                    warn("The docker build command is slower than expected. You may increase performance by adding " + 
+                        "unneeded files and directories such as any Liberty runtime directories to a .dockerignore file in " +
+                        dockerContext.getCanonicalPath()+ ".");
+                } catch (IOException e) {
+                    error("Exception while retrieving the path name of the docker context directory", e);
```
ericglau, line 1173: Instead of just giving an error here, perhaps use dockerContext.getAbsolutePath() to get the path if an IOException occurs and still give the actual "docker build command is slow" warning.

e.g. something like
String buildContext;
try {
buildContext = dockerContext.getCanonicalPath();
} catch (IOException e) {
buildContext = dockerContext.getAbsolutePath();
}
warn(......buildContext);

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1149,6 +1153,29 @@ private String execDockerCmd(String command, int timeout) {
...
+    private void checkDockerIgnore(long startTime) {
+        if (System.currentTimeMillis() - startTime < DOCKER_BUILD_SOFT_TIMEOUT) {
+            return;
+        }
+        File dockerfileToUse = dockerfile != null ? dockerfile : defaultDockerfile;
+        if (dockerfileToUse.exists()) {
+            File dockerContext = dockerfileToUse.getParentFile();
+            debug("checkDockerIgnore, dockerContext="+dockerContext.getAbsolutePath());
+            File dockerIgnore = new File(dockerContext, ".dockerignore");
+            if (!dockerIgnore.exists()) { // provide some advice
```
TrevCraw, line 1167: Should we suggest adding the runtime directories to `.dockerignore` even if they already have a `.dockerignore` file? I suppose this could get annoying if they already have the runtime dirs in `.dockerignore`, but if they don't, this would still be good advice.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: Not a bad idea.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: I think if they are using a `.dockerignore` they already know to search around for large directories to add to the file.

PR #185, opened 2020-09-16
Add hotkey to restart server, add formatting to dev mode startup messages

ericglau:
Part of https://github.com/OpenLiberty/ci.maven/issues/927

Add a hotkey where you type 'r' and press Enter to restart the server.  If using container mode, this will rebuild the image and restart the container instead.

Update dev mode startup messages to include this.

Update Dockerfile warnings to suggest this hotkey for the COPY commands that are currently not supported for hot deployment.

Adds formatting to dev mode startup messages as follows:
```
[INFO] ************************************************************************
[INFO] *    Liberty dev mode has started!
[INFO] *        To run tests on demand, press Enter.
[INFO] *        To rebuild the Docker image and restart the container, type 'r' and press Enter.
[INFO] *        To stop the server and quit dev mode, press Ctrl-C or type 'q' and press Enter.
[INFO] ************************************************************************
```
or the below for non-container:
```
[INFO] ************************************************************************
[INFO] *    Liberty dev mode has started!
[INFO] *        To run tests on demand, press Enter.
[INFO] *        To restart the server, type 'r' and press Enter.
[INFO] *        To stop the server and quit dev mode, press Ctrl-C or type 'q' and press Enter.
[INFO] ************************************************************************
```

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -855,7 +855,7 @@ public abstract class DevUtil {
                         "'. There must be at least two arguments for the COPY command, a source path and a destination path.");
                     }
                     if (line.contains("$")) {
-                        warn("The Dockerfile line '" + line + "' will not be able to be hot deployed to the dev mode container. Dev mode does not currently support environment variables in COPY commands.");
+                        warn("The Dockerfile line '" + line + "' will not be able to be hot deployed to the dev mode container. Dev mode does not currently support environment variables in COPY commands. If you make changes to files specified by this line, type 'r' and press Enter to rebuild the Docker image and restart the container.");
                         continue;
                     }
                     List<String> srcOrDestArguments = new ArrayList<String>();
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2238,9 +2266,15 @@ private void processJavaCompilation(File outputDirectory, File testOutputDirecto
         }
     }
  
-    private void checkStopDevMode() throws PluginScenarioException {
+    private void checkStopDevMode(boolean skipOnRestart) throws PluginScenarioException {
         // stop dev mode if the server has been stopped by another process
         if (serverThread == null || serverThread.getState().equals(Thread.State.TERMINATED)) {
+            // server is restarting if devStop was set to true and we have not called the shutdown hook
+            boolean restarting = devStop.get() && !calledShutdownHook.get();
```
TrevCraw, line 2273: I noticed that just `devStop` is used here, but we use `this.devStop` below. Is there a reason to use `this`? If not, which is the better practice? We should probably be consistent.

ericglau: Good catch, `this` should not be necessary.

TrevCraw (approved):

turkeylurkey (approved):

TrevCraw (approved):

turkeylurkey (approved):

PR #187, opened 2020-09-21
Display stdout and stderr when running docker build just like docker run

turkeylurkey:
I moved some code around so the diffs look a little bigger.
Also, docker build goes into "console mode" when you redirect the output rather than display it on a terminal.
Fixes OpenLiberty/ci.maven#925

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -951,11 +951,49 @@ public abstract class DevUtil {
             debug("Docker build context: " + userDockerfile.getParent());
             buildCmd = "docker build -f " + tempDockerfile + " -t " + imageName + " " + userDockerfile.getParent();
             info(buildCmd);
-            execDockerCmd(buildCmd, dockerBuildTimeout);
+            long startTime = System.currentTimeMillis();
+            runCommandAndLog(buildCmd, dockerBuildTimeout);
+            checkDockerIgnore(startTime, userDockerfile);
             info("Completed building Docker image.");
+        } catch (IllegalThreadStateException  e) {
+            // the timeout was too short and the docker command has not yet completed. There is no exit value.
+            debug("IllegalThreadStateException, message="+e.getMessage());
+            error("The docker build command did not complete within the timeout period: " + dockerBuildTimeout + " seconds. " +
+                "Use the dockerBuildTimeout option to specify a longer period or " +
+                "add files not needed in the container to the .dockerignore file.", e);
+            throw new RuntimeException("The docker command did not complete within the timeout period: " + dockerBuildTimeout + " seconds. ");
+        } catch (IOException e) {
+            error("Error building container: " + e.getMessage());
+            throw new RuntimeException(e);
+        } catch (InterruptedException e) {
+            error("Thread was interrupted while building the container: " + e.getMessage());
         } catch (RuntimeException r) {
             error("Error building Docker image: " + r.getMessage());
             throw new PluginExecutionException("Could not build Docker image using Dockerfile: " + userDockerfile.getAbsolutePath() + ". Address the following docker build error and then start dev mode again: " + r.getMessage(), r);
+        }
+    }
+
+    // Suggest a performance improvement if docker build takes too long.
+    private static final long DOCKER_BUILD_SOFT_TIMEOUT = 30000;
+    private void checkDockerIgnore(long startTime, File userDockerfile) {
+        if (System.currentTimeMillis() - startTime < DOCKER_BUILD_SOFT_TIMEOUT) {
+            return;
+        }
+        if (userDockerfile.exists()) {
+            File dockerContext = userDockerfile.getParentFile();
+            debug("checkDockerIgnore, dockerContext="+dockerContext.getAbsolutePath());
+            File dockerIgnore = new File(dockerContext, ".dockerignore");
+            if (!dockerIgnore.exists()) { // provide some advice
+                String buildContext;
+                try {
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -951,14 +951,55 @@ private void buildDockerImage(File tempDockerfile, File userDockerfile) throws P
...
+            info("Completed building Docker image.");
+        } catch (IllegalThreadStateException  e) {
+            // the timeout was too short and the docker command has not yet completed.
+            debug("IllegalThreadStateException, message="+e.getMessage());
+            error("The docker build command did not complete within the timeout period: " + dockerBuildTimeout + " seconds. " +
+                "Use the dockerBuildTimeout option to specify a longer period or " +
+                "add files not needed in the container to the .dockerignore file.", e);
+            throw new PluginExecutionException("Could not build Docker image using Dockerfile: " + 
+                userDockerfile.getAbsolutePath() + ". Address the following docker build error and then start dev mode again: " +
+                "The docker build command did not complete within the timeout period: " + dockerBuildTimeout + " seconds", e);
```
TrevCraw, line 966: I think we should remove the `Address the following docker build error and then start dev mode again:` piece, because we are not showing the docker build error here.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -951,14 +951,55 @@ private void buildDockerImage(File tempDockerfile, File userDockerfile) throws P
...
+        } catch (IOException e) {
+            error("Input or output error building Docker image: " + e.getMessage());
+            throw new RuntimeException(e);
+        } catch (InterruptedException e) {
+            error("Thread was interrupted while building the Docker image: " + e.getMessage());
+            throw new PluginExecutionException("Could not build Docker image using Dockerfile: " + userDockerfile.getAbsolutePath() + ". Address the following docker build error and then start dev mode again: " + e.getMessage(), e);
```
TrevCraw, line 972: Is it the same as above here? Would `e` be the docker build error or some other Exception?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: Good point. If anyone were to ask why we are duplicating all these messages here I would respond it is because we want to tune the response to the situation. In this case a docker build. I'll take a look.

ericglau, line 966: Is there a benefit of providing two different messages for `error` and `PluginExecutionException` (compared to making them the same message, or just combining them into the PluginExecutionException)?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1006,20 +1020,49 @@ public void run() {
...
-    private String runCmd(String cmd) throws IOException, InterruptedException {
-        return runCmd(true, cmd);
+    private Process getRunProcess(String command) throws IOException {
+        ProcessBuilder processBuilder = new ProcessBuilder();
+        processBuilder.command(getCommandTokens(command));
+        return processBuilder.start();
     }
 
-    private String runCmd(boolean report, String cmd) throws IOException, InterruptedException {
+    private void runCommandAndLog(final Process startingProcess, int timeout) throws InterruptedException {
```
ericglau, line 1029: Should this be renamed to something like `runDockerCommandAndLog`?  Although it's not related to Docker except for its debug message string.

Perhaps rename this method and/or `execDockerCmd` method to avoid confusion, since both method names sound like they do the same thing.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: My thinking is that the first error message has all the details and when you review it you see it in context. The exception message is seen at the end so I made it more general. It includes the docker file that you might have seen in the context of the original message. I agree that it is quite wordy as it is. Perhaps in the case of a plugin exception we're going to halt execution so the distinction I just made is not so important.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: How about execDockerCmd() and execDockerCmdAndLog()? They do the same thing except the one that logs?

ericglau: That's fine.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: Since we are throwing PluginExecutionException dev mode will exit so the error message is not necessary. We'll just put the message in the exception.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: I just wanted to capture the interrupted exception in case that was relevant. This is better done with a debug message and then the standard exit.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: Combined the messages into the exception only.

ericglau (approved):

PR #188, opened 2020-09-23
When startContainer() fails, check for and remove the left over container.

turkeylurkey:
Also improved the stopContainer() not to try to remove the container when it is not found.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1019,11 +1019,16 @@ public abstract class DevUtil {
         } catch (InterruptedException e) {
             error("Thread was interrupted while starting the container: " + e.getMessage());
         } catch (RuntimeException r) {
-            // remove container in case of an error trying to run the container because the docker run --rm will not rm the container
-            String containerId = getContainerId();
-            if (containerId != null && !containerId.isEmpty()) {
-                String dockerRmCmd = "docker container rm " + containerId;
-                execDockerCmd(dockerRmCmd, 10);
+            try {
+                // remove container in case of an error trying to run the container because the docker run --rm will not rm the container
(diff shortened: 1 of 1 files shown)


ericglau (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1018,6 +1018,14 @@ private void startContainer() {
             throw new RuntimeException(e);
         } catch (InterruptedException e) {
             error("Thread was interrupted while starting the container: " + e.getMessage());
+        } catch (RuntimeException r) {
+            // remove container in case of an error trying to run the container because the docker run --rm will not rm the container
+            String containerId = getContainerId();
+            if (containerId != null && !containerId.isEmpty()) {
+                String dockerRmCmd = "docker container rm " + containerId;
+                execDockerCmd(dockerRmCmd, 10);
```
ericglau, line 1026: If the remove command fails (if the container somehow was already removed), should we catch the exception from execDockerCmd and suppress it?

turkeylurkey: If that happens then the docker run failure is replaced by the docker rm failure:
```
[ERROR] Failed to execute goal io.openliberty.tools:liberty-maven-plugin:3.3-M4-SNAPSHOT:devc (default-cli) on project demo-devmode-maven: An error occurred while starting the container: Error: No such container: 0531cc9908b8d RC=1 -> [Help 1]
```
Should we surround the docker rm with a try/catch to throw away that error and report the docker run error?

turkeylurkey: I asked the same question as you. I think it would be better if we suppress the second error. We need to report the actual user error.

ericglau (approved):

PR #191, opened 2020-09-25
854 multi instance support

TrevCraw:
Part of https://github.com/OpenLiberty/ci.maven/issues/854


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1194,7 +1192,7 @@ private String readStdOut(Process p) throws IOException, InterruptedException {
         String line;
         StringBuffer allLines = new StringBuffer();
         while ((line = in.readLine())!= null) {
-            allLines.append(line);
+            allLines.append(line + " ");
```
turkeylurkey, line 1195: We are using StringBuffer to avoid using concatenation. Could you make this .append(line).append(" ")?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1210,21 +1208,25 @@ private String readStdOut(Process p) throws IOException, InterruptedException {
...
-            command.append(" -p "+httpPort+":"+httpPort);
-        } else {
-            command.append(" -p 9080:9080");
-        }
-        if (httpsPort != null) {
-            command.append(" -p "+httpsPort+":"+httpsPort);
-        } else {
-            command.append(" -p 9443:9443");
+        if (!skipDefaultPorts) {
+            if (httpPort != null) {
```
turkeylurkey, line 1212: Do we still want to use the http port detected in the server? This code is strange because the first time we run httpPort==null and we use 9080 but if the server config specifies port 9081 and we restart the container we will "default" to 9081 instead of 9080. It is more correct but not expected.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1210,21 +1208,25 @@ private String readStdOut(Process p) throws IOException, InterruptedException {
...
+                command.append(" -p "+httpPort+":"+httpPort);
+            } else {
+                command.append(" -p 9080:9080");
+            }
+            if (httpsPort != null) {
```
turkeylurkey, line 1217: Same comment as line 1212.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: Yeah, I am not sure the best way to approach this. I was hoping to have a discussion about how we should handle the `httpPort` and `httpsPort` variables.  Because here we are talking about the Liberty server, so the ports are inside of the container, but I believe these variables are used for the integration tests, so they should actually be associated with the local ports.

I opened an issue to address this issue plus avoiding default port conflicts: https://github.com/OpenLiberty/ci.maven/issues/946
If you guys think the `httpPort` and `httpsPort` variables situation deserves its own issue, feel free to separate.

turkeylurkey: After the server starts we detect these ports so they are set properly for the test phase but they are not ready for the docker run step. Except when we restart. Unless the user updates the server config port to 9082 and restarts in which case we would publish 9081 and use 9082 for the tests. Then if there is a second restart we would publish 9082. The -p option is always one restart behind.

turkeylurkey (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1262,6 +1266,41 @@ private String getContainerCommand() {
...
+        }
+        String[] containerNames = result.split(" ");
+        int highestNum = -1;
+        for(int i = 0; i < containerNames.length; i++) {
+            String name = containerNames[i];
+            int num = -1;
+            if (name.equals(DEVMODE_CONTAINER_BASE_NAME)) {
+                num = 0;
+            }
+            else if (name.startsWith(DEVMODE_CONTAINER_BASE_NAME + "_")) {
```
ericglau, line 1284: Use `} else if ` for consistency with the rest of the file.

ericglau (approved):

PR #198, opened 2020-10-09
Add ability to resolve env vars for an include file from bootstrap

jjvilleg:
Resolves https://github.com/OpenLiberty/ci.maven/issues/957

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
@@ -409,8 +409,8 @@ public abstract class ServerFeatureUtil {
             if(!VALID_LIBERTY_DIRECTORY_PROPERTY.contains(envDirectoryProperty)) {
                 // Check if property is a reference to a configured bootstrap property
                 String bootStrapValue = properties.getProperty(envDirectoryProperty, "\\$\\{" + envDirectoryProperty + "\\}");
-                if(boostrapValue != null) {
-                    m.appendReplacement(sb, removeEncapsulatingEnvVarSyntax(boostrapValue);
+                if(bootStrapValue != null) {
+                    m.appendReplacement(sb, removeEncapsulatingEnvVarSyntax(bootStrapValue, properties));
                 }
                 else {
                     warn("The directory property " + envDirectoryProperty + "specified is not a predifined Liberty directory property or a configured boostrap property.");
--- a/src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
@@ -56,6 +56,8 @@ public abstract class ServerFeatureUtil {
     public static final String REPOSITORY_RESOLVER_ARTIFACT_ID = "repository-resolver";
     public static final String INSTALL_MAP_ARTIFACT_ID = "install-map";
     private static final int COPY_FILE_TIMEOUT_MILLIS = 5 * 60 * 1000;
+    private static final List<String> VALID_LIBERTY_DIRECTORY_PROPERTY = Arrays.asList("wlp.install.dir", "wlp.user.dir", "usr.extension.dir", 
+            "shared.app.dir", "shared.config.dir", "shared.resource.dir", "shared.stackgroup.dir", "server.config.dir", "server.output.dir");
     
     /**
      * Log debug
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -380,18 +382,39 @@ private Properties getBootstrapProperties(File bootstrapProperties) {
     private String evaluateExpression(Properties properties, String expression) {
         String value = expression;
         if (expression != null) {
-            Pattern p = Pattern.compile("\\$\\{([^\\}]*)\\}");
+            Pattern p = Pattern.compile("\\$\\{(.*?)\\}");
```
cherylking, line 385: I think this will cause a problem if the expression passed in contains two variables like the following:

${var1}/${var2}

That initial pattern was supposed to do a non greedy match I think and find the first variable, which would return `var1` at line 389. After your change to the pattern, I think it will return `var1/${var2`.

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -380,18 +382,39 @@ private Properties getBootstrapProperties(File bootstrapProperties) {
...
             }
             m.appendTail(sb);
             value = sb.toString();
         }
         return value;
     }
-    
+
+    private String removeEncapsulatingEnvVarSyntax(String propertyValue){
+        Pattern p = Pattern.compile("\\$\\{(.*?)\\}");
```
cherylking, line 404: same issue as I described in evaluateExpression method

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -380,18 +382,39 @@ private Properties getBootstrapProperties(File bootstrapProperties) {
...
+        Matcher m = p.matcher(propertyValue);
+        StringBuffer sb = new StringBuffer();
+        while (m.find()) {
+            String envDirctoryProperty = m.group(1);
```
cherylking, line 408: nit...variable spelled wrong...envDirectoryProperty

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -380,18 +382,39 @@ private Properties getBootstrapProperties(File bootstrapProperties) {
...
+            if(!VALID_LIBERTY_DIRECTORY_PROPERTY.contains(envDirctoryProperty)) {
+                warn("The directory property " + envDirctoryProperty + "specified is not a predifined Liberty directory property");
+                return null;
+            }
+            else {
+                m.appendReplacement(sb, envDirctoryProperty);
```
cherylking, line 414: What if the referenced variable is another bootstrap property? Probably should handle resolving that also.

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
cherylking: The change is fine. I see that the ? makes it non-greedy.

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -380,18 +382,46 @@ private Properties getBootstrapProperties(File bootstrapProperties) {
...
+        while (m.find()) {
+            String envDirectoryProperty = m.group(1);
+            if(!VALID_LIBERTY_DIRECTORY_PROPERTY.contains(envDirectoryProperty)) {
+                // Check if property is a reference to a configured bootstrap property
+                String bootStrapValue = properties.getProperty(envDirectoryProperty, "\\$\\{" + envDirectoryProperty + "\\}");
+                if(bootStrapValue != null) {
+                    m.appendReplacement(sb, removeEncapsulatingEnvVarSyntax(bootStrapValue, properties));
+                }
+                else {
+                    warn("The directory property " + envDirectoryProperty + "specified is not a predifined Liberty directory property or a configured boostrap property.");
```
cherylking, line 416: I would take out the first reference to directory in this log message....change to 

"The referenced property " +envDirectoryProperty + "is not a predefined Liberty directory property or a configured bootstrap property."

cherylking (changes requested):
Just the one suggestion.

cherylking (approved):

PR #206, opened 2020-10-20
Track directories in Dockerfile COPY commands

ericglau:
Fixes https://github.com/OpenLiberty/ci.maven/issues/846


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -323,6 +324,10 @@ public abstract boolean recompileBuildFile(File buildFile, List<String> artifact
     protected List<String> srcMount = new ArrayList<String>();
     protected List<String> destMount = new ArrayList<String>();
     private boolean firstStartup = true;
+    private Set<Path> dockerfileDirectoriesToWatch = new HashSet<Path>();
+    private Set<Path> dockerfileDirectoriesChildren = new HashSet<Path>();
```
turkeylurkey, line 328: As discussed, perhaps a better name for this field.

turkeylurkey (approved):

TrevCraw (approved):

PR #207, opened 2020-10-22
Pass in Liberty directory property files

cherylking:
For OpenLiberty/ci.gradle#551


jjvilleg (approved):

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -108,6 +108,19 @@
      */
     public Set<String> getServerFeatures(File serverDirectory) {
```
jjvilleg, line 109: Looks good, but just a note. I didn't see any calls to this method anymore in the gradle pr. Is this being left in for maven?

ericglau (approved):

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
cherylking: Yes, left in for maven. I'm going to look at maven next to see if it needs to change also. I'll remove this method when/if maven no longer uses it.

PR #208, opened 2020-10-26
Handle Dockerfile ADD commands

turkeylurkey:
Fixes https://github.com/OpenLiberty/ci.maven/issues/962
Also fixes ConcurrentModificationException when registering multiple dockerfileDirectoriesToWatch in the watchFiles loop.
Do we need to build a list of directories registered in case there is an exception? Then remove the list instead of dockerfileDirectoriesToWatch.clear()? It looks like if there *is* an exception then dev mode exits.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -843,7 +843,7 @@ public abstract class DevUtil {
                 // The command must be to the left of any comments.
                 String[] cmdSegments = trimLine.split("#")[0].split("\\s+");
                 // if the line starts with COPY and the second to last segment ends with ".war", it is a WAR file COPY line
-                if (cmdSegments[0].equalsIgnoreCase("COPY")) {
+                if (cmdSegments[0].equalsIgnoreCase("COPY") || cmdSegments[0].equalsIgnoreCase("ADD")) {
                     if (cmdSegments.length < 3) {
                         throw new PluginExecutionException("Incorrect syntax on this line in the Dockerfile: '" + line + 
                         "'. There must be at least two arguments for the COPY command, a source path and a destination path.");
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -843,7 +843,7 @@ protected void removeWarFileLines(List<String> dockerfileLines) throws PluginExe
                 // The command must be to the left of any comments.
                 String[] cmdSegments = trimLine.split("#")[0].split("\\s+");
                 // if the line starts with COPY and the second to last segment ends with ".war", it is a WAR file COPY line
-                if (cmdSegments[0].equalsIgnoreCase("COPY")) {
+                if (cmdSegments[0].equalsIgnoreCase("COPY") || cmdSegments[0].equalsIgnoreCase("ADD")) {
                     if (cmdSegments.length < 3) {
                         throw new PluginExecutionException("Incorrect syntax on this line in the Dockerfile: '" + line + 
                         "'. There must be at least two arguments for the COPY command, a source path and a destination path.");
```
ericglau, line 849: The exception message should be updated as well.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -887,8 +887,8 @@ protected void processCopyLines(List<String> dockerfileLines, String buildContex
                 // Break the Dockerfile line down into segments based on any amount of whitespace.
                 // The command must be to the left of any comments.
                 String[] cmdSegments = trimLine.split("#")[0].split("\\s+");
-                // If the line starts with COPY
-                if (cmdSegments[0].equalsIgnoreCase("COPY")) {
+                // If the line starts with COPY or ADD
+                if (cmdSegments[0].equalsIgnoreCase("COPY") || cmdSegments[0].equalsIgnoreCase("ADD")) {
```
ericglau, line 891: Update the exceptions and warning messages in this method.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: ok

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: ok

ericglau (approved):

PR #215, opened 2020-11-02
Detect when the server http/s ports are used and try the next port in sequence

turkeylurkey:
Fixes https://github.com/OpenLiberty/ci.maven/issues/947
I duplicated the message and added ‘<‘ to the second copy because of the semantics indicated by the method formatAttentionMessage() which suggests padding could be added to the beginning or the end. Hopefully in the future these messages will be translated and the separate messages should be easier to translate.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -120,6 +120,9 @@ public abstract class DevUtil {
             "___jb_tmp___", "___jb_old___" };
 
     private static final String[] DEFAULT_COMPILER_OPTIONS = new String[] { "-g", "-parameters" };
+    private static final int LIBERTY_DEFAULT_HTTP_PORT = 9080;
+    private static final int LIBERTY_DEFAULT_HTTPS_PORT = 9443;
+    private static final int LIBERTY_DEFAULT_DEBUG_PORT = 7777;
 
     /**
      * Log debug
@@ -1258,14 +1261,14 @@ public abstract class DevUtil {
         if (!skipDefaultPorts) {
             int httpPortToUse, httpsPortToUse;
             try {
-                httpPortToUse = findAvailablePort(9080, false);
-                httpsPortToUse = findAvailablePort(9443, false);
+                httpPortToUse = findAvailablePort(LIBERTY_DEFAULT_HTTP_PORT, false);
+                httpsPortToUse = findAvailablePort(LIBERTY_DEFAULT_HTTPS_PORT, false);
             } catch (IOException x) {
-                httpPortToUse = 9080;
-                httpsPortToUse = 9443;
+                httpPortToUse = LIBERTY_DEFAULT_HTTP_PORT;
+                httpsPortToUse = LIBERTY_DEFAULT_HTTPS_PORT;
             }
-            command.append(" -p ").append(httpPortToUse).append(":9080");
-            command.append(" -p ").append(httpsPortToUse).append(":9443");
+            command.append(" -p ").append(httpPortToUse).append(":").append(LIBERTY_DEFAULT_HTTP_PORT);
+            command.append(" -p ").append(httpsPortToUse).append(":").append(LIBERTY_DEFAULT_HTTPS_PORT);
         }
         
         if (libertyDebug) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1256,7 +1256,16 @@ private String readStdOut(Process p) throws IOException, InterruptedException {
...
         StringBuilder command = new StringBuilder("docker run --rm");
         if (!skipDefaultPorts) {
-            command.append(" -p 9080:9080 -p 9443:9443");
+            int httpPortToUse, httpsPortToUse;
+            try {
+                httpPortToUse = findAvailablePort(9080, false);
+                httpsPortToUse = findAvailablePort(9443, false);
+            } catch (IOException x) {
+                httpPortToUse = 9080;
+                httpsPortToUse = 9443;
```
ericglau, line 1265: We should also log the error/exception here.

turkeylurkey: ok

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1962,29 +1965,45 @@ private void printDevModeMessages(boolean inputUnavailable, boolean startup) {
         }
         if (startup) {
             if (container) {
+                boolean nonDefaultHttpPortUsed = !skipDefaultPorts && !String.valueOf(LIBERTY_DEFAULT_HTTP_PORT).equals(httpPort);
+                boolean nonDefaultHttpsPortUsed = !skipDefaultPorts && !String.valueOf(LIBERTY_DEFAULT_HTTPS_PORT).equals(httpsPort);
+                int debugPort = (alternativeDebugPort == -1 ? libertyDebugPort : alternativeDebugPort);
+                boolean nonDefaultDebugPortUsed = !skipDefaultPorts && LIBERTY_DEFAULT_DEBUG_PORT != debugPort;
```
ericglau, line 1972: `skipDefaultPorts` should not affect this boolean, since it is not related to the debug port.  Perhaps it should check if `libertyDebug` is true instead .

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1264,6 +1264,7 @@ private String getContainerCommand() {
                 httpPortToUse = findAvailablePort(LIBERTY_DEFAULT_HTTP_PORT, false);
                 httpsPortToUse = findAvailablePort(LIBERTY_DEFAULT_HTTPS_PORT, false);
             } catch (IOException x) {
+                debug("IOException trying to findAvailablePort(), using default ports.");
```
ericglau, line 1267: I was thinking to actually print a user-facing error() and include the exception message.  Perhaps even include the stacktrace in the same error or as a separate debug log statement.  Because if findAvailablePort gives an IOException, then that means either there was no available port, or there was an error when opening a socket.  Either seems like a severe issue, and if the `docker run` command fails afterwards (due to those issues), the exception can help to diagnose the problem.

turkeylurkey: Agreed.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey:
```
[INFO] *    Liberty container port information:
[WARNING] *        The Liberty container is using non-default host ports.
[INFO] *        Internal container HTTP port [ 9080 ] is mapped to Docker host port [ 9080 ]
[INFO] *        Internal container HTTPS port [ 9443 ] is mapped to Docker host port [ 9443 ]
[INFO] *        Liberty debug port mapped to Docker host port: [ 60141 ] <
[INFO] ************************************************************************
```

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1256,7 +1259,17 @@ private String readStdOut(Process p) throws IOException, InterruptedException {
     private String getContainerCommand() {
         StringBuilder command = new StringBuilder("docker run --rm");
         if (!skipDefaultPorts) {
-            command.append(" -p 9080:9080 -p 9443:9443");
+            int httpPortToUse, httpsPortToUse;
+            try {
+                httpPortToUse = findAvailablePort(LIBERTY_DEFAULT_HTTP_PORT, false);
+                httpsPortToUse = findAvailablePort(LIBERTY_DEFAULT_HTTPS_PORT, false);
+            } catch (IOException x) {
+                error("An error occurred while trying to find a free network port, using default port numbers.", x);
```
ericglau, line 1267: `error("An error occurred while trying to find an available network port. Using default port numbers.", x);`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1937,29 +1966,45 @@ private void printDevModeMessages(boolean inputUnavailable, boolean startup) {
...
         if (startup) {
             if (container) {
+                boolean nonDefaultHttpPortUsed = !skipDefaultPorts && !String.valueOf(LIBERTY_DEFAULT_HTTP_PORT).equals(httpPort);
+                boolean nonDefaultHttpsPortUsed = !skipDefaultPorts && !String.valueOf(LIBERTY_DEFAULT_HTTPS_PORT).equals(httpsPort);
+                boolean nonDefaultDebugPortUsed = alternativeDebugPort != -1; // this is set when a random ephemeral port is selected
                 if (containerHttpPort != null || containerHttpsPort != null || libertyDebug) {
                     info(formatAttentionMessage(""));
                     info(formatAttentionTitle("Liberty container port information:"));
                 }
+                if (nonDefaultHttpPortUsed || nonDefaultHttpsPortUsed || nonDefaultDebugPortUsed) {
```
ericglau, line 1976: This is fine with the way the variables are currently set up, but as a precaution (against possible bugs if we change the code later on), it would be safer to also check if libertyDebug is enabled.  e.g.
```
if (nonDefaultHttpPortUsed || nonDefaultHttpsPortUsed || (libertyDebug && nonDefaultDebugPortUsed)) {
```

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: 😎👍

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: 😎👍

ericglau (approved):

PR #219, opened 2020-11-06
Check if the docker run command is still running before trying to stop it

turkeylurkey:
Fixes https://github.com/OpenLiberty/ci.maven/issues/960

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1174,14 +1174,12 @@ public abstract class DevUtil {
 
     private void stopContainer() {
         try {
-            if (dockerRunProcess != null) {
-                // see if docker run command (container) is still running before trying to stop it.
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1175,10 +1175,13 @@ private String copyStreamToBuildLog(InputStream stream, boolean info) {
     private void stopContainer() {
         try {
             if (dockerRunProcess != null) {
-                info("Stopping container...");
-                String dockerStopCmd = "docker stop " + containerName;
-                debug("Stopping container " + containerName);
-                execDockerCmd(dockerStopCmd, 30);
+                // see if docker run command (container) is still running before trying to stop it.
+                if (dockerRunProcess.isAlive()) {
```
ericglau, line 1179: Could combine these two if-statements into one with &&

ericglau (approved):

PR #220, opened 2020-11-09
Add keepTempDockerfile option

turkeylurkey:
Used in two places: control tempDockerfile.deleteOnExit() and in cleanUpTempDockerfile() called from shutdown hook.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1015,7 +1015,9 @@ public abstract class DevUtil {
             tempDockerfile = File.createTempFile("tempDockerfile", "");
             debug("temp Dockerfile: " + tempDockerfile);
             tempDockerfilePath = tempDockerfile.toPath(); // save name to clean up later
-            if (!keepTempDockerfile) {
+            if (keepTempDockerfile) {
+                info("Saving temporary Dockerfile named:"+tempDockerfilePath);
+            } else {
                 // set the tempDockerfile to be deleted when the JVM exits
                 tempDockerfile.deleteOnExit();
             }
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1012,9 +1014,11 @@ protected File prepareTempDockerfile(File dockerfile) throws PluginExecutionExce
             debug("Creating temp Dockerfile...");
             tempDockerfile = File.createTempFile("tempDockerfile", "");
             debug("temp Dockerfile: " + tempDockerfile);
-            tempDockerfilePath = tempDockerfile.toPath();
-            // set the tempDockerfile to be deleted when the JVM exits
-            tempDockerfile.deleteOnExit();
+            tempDockerfilePath = tempDockerfile.toPath(); // save name to clean up later
```
ericglau, line 1017: If `keepTempDockerfile` is enabled, it would also be useful to print out an `info` message on what the temp Dockerfile path is.

ericglau (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: Ok, that is easier than finding the name in the docker build command.

turkeylurkey: [INFO] Saving temporary Dockerfile named:/var/folders/p1/v7svjg7s6xngg377xtsjxcmc0000gn/T/tempDockerfile7677461533838714336

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1012,9 +1014,13 @@ protected File prepareTempDockerfile(File dockerfile) throws PluginExecutionExce
...
+            if (keepTempDockerfile) {
+                info("Saving temporary Dockerfile named:"+tempDockerfilePath);
```
ericglau, line 1019: Perhaps use terminology in line with the option name so it's easier to find.  e.g.:
```
info("Keeping temporary Dockerfile: " + tempDockerfilePath);
```

turkeylurkey: Flashy! Done.

PR #236, opened 2020-12-08
Set DOCKER_BUILDKIT=0 environment var for docker build and run on Mac and Windows

turkeylurkey:
Fixes https://github.com/OpenLiberty/ci.maven/issues/1045
Also increase docker command timeouts from 10s to 20s except server stop from 30s to 40s. VMs can be slower than physical machines.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -126,6 +126,7 @@ public abstract class DevUtil {
     private static final int LIBERTY_DEFAULT_HTTP_PORT = 9080;
     private static final int LIBERTY_DEFAULT_HTTPS_PORT = 9443;
     private static final int LIBERTY_DEFAULT_DEBUG_PORT = 7777;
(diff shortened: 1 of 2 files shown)


TrevCraw (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1127,6 +1128,13 @@ private void startContainer() {
     private Process getRunProcess(String command) throws IOException {
         ProcessBuilder processBuilder = new ProcessBuilder();
         processBuilder.command(getCommandTokens(command));
+        if (!System.getProperty("os.name").equalsIgnoreCase("linux")){
```
ericglau, line 1131: There is an OSUtil.isWindows().  You could move this check to OSUtils as isLinux(), or just use isWindows().  Depending on whether Mac needs this case or not.

turkeylurkey: I like this idea, we use the check many times. The same failure exists on Mac but I don't know how to create a MacOS VM.

PR #237, opened 2020-12-10
Install features to container

ericglau:
Part of https://github.com/OpenLiberty/ci.maven/issues/811

- Increase docker build timeout to 600 seconds.
- If feature not found messages occur, give a warning suggesting to add RUN features.sh
- Install features using featureUtility and mounted Maven cache (but skip during dev mode startup and container rebuilds)

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -92,7 +92,7 @@ import io.openliberty.tools.common.plugins.config.ServerConfigDropinXmlDocument;
 /**
  * Utility class for dev mode.
  */
-public abstract class DevUtil {
+public abstract class DevUtil extends AbstractContainerSupportUtil {
 
     private static final String START_SERVER_MESSAGE_PREFIX = "CWWKF0011I:";
     private static final String START_APP_MESSAGE_REGEXP = "CWWKZ0001I.*";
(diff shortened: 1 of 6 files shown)


turkeylurkey (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1200,14 +1203,17 @@ private String copyStreamToBuildLog(InputStream stream, boolean info) {
...
-                        }
+                            "Java classes were compiled with a higher version of Java than the JVM in the container. To resolve this issue, set the source and target Java versions in your Maven build to correspond to the Java version used in your Dockerfile or its parent image, then clean the project output and restart dev mode.",
+                            false);
+
+                    // Look for features not available
+                    String errMsg = "Feature definitions were not found in the container. To install features to the container, specify 'RUN features.sh' in your Dockerfile. For an example of how to configure a Dockerfile, see https://github.com/OpenLiberty/ci.docker";
+                    if (!shownFeaturesShWarning) {
+                        boolean alerted = alertOnServerError(line, "CWWKF0001E", errMsg, errMsg, true);
+                        shownFeaturesShWarning = alerted;
```
TrevCraw, line 1216: Just curious why setting `shownFeaturesShWarning` was done in two steps instead of one. Is this better practice?

TrevCraw (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: Good catch. Fixed.

PR #240, opened 2020-12-14
Check the image name for compliance with Docker rules.

turkeylurkey:
Fixes https://github.com/OpenLiberty/ci.maven/issues/1058


ericglau (approved):
We should probably also change https://github.com/OpenLiberty/ci.maven/blob/master/liberty-maven-plugin/src/main/java/io/openliberty/tools/maven/server/DevMojo.java#L291 to return `project.getArtifactId()` instead.  I still agree we should sanitize the input here since even the artifactId is not guaranteed to satisfy Docker rules, and Gradle still uses the project name (since there is not really an artifactId concept in basic Gradle projects).

PR #241, opened 2020-12-15
Fix when to display features warning message

ericglau:
Fixes https://github.com/OpenLiberty/ci.maven/issues/1061

- Only display RUN features.sh warning message if all of these conditions are met: server still is starting up, Dockerfile does not have `RUN features.sh`, and the message has not been displayed before during this startup session.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -339,9 +339,9 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
     private Set<WatchKey> dockerfileDirectoriesWatchKeys = new HashSet<WatchKey>();
     private Set<FileAlterationObserver> dockerfileDirectoriesFileObservers = new HashSet<FileAlterationObserver>();
     private final JavaCompilerOptions compilerOptions;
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -924,6 +931,25 @@ protected void disableOpenJ9SCC(List<String> dockerfileLines) {
         }
     }
 
+    protected void detectFeaturesSh(List<String> dockerfileLines) {
+        // Reset features.sh warning flag
+        shownFeaturesShWarning.set(false);
+
+        final String FEATURES_SH_COMMAND_LOWERCASE = "run features.sh";
```
turkeylurkey, line 938: I dont think getCleanedLines() removes internal whitespace in the line so if the person types RUN<blank><blank>features.sh then the check should fail. Can we use a regex here?

turkeylurkey (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: This is also not handled when detecting `RUN configure.sh`.  It may be better to change getCleanedLines() to remove internal whitespace. I'll open a separate issue.

PR #245, opened 2021-01-05
Update slow docker build message to account for RUN features.sh command.

turkeylurkey:
Basically two warnings (possibly) in one message because they are related. 
Fixes: https://github.com/OpenLiberty/ci.maven/issues/1069

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1102,7 +1102,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
             info(buildCmd);
             long startTime = System.currentTimeMillis();
             execDockerCmdAndLog(getRunProcess(buildCmd), dockerBuildTimeout);
-            checkDockerIgnore(startTime, userDockerfile.getParentFile());
+            checkDockerBuildTime(startTime, userDockerfile.getParentFile());
             info("Completed building Docker image.");
         } catch (IllegalThreadStateException  e) {
             // the timeout was too short and the docker command has not yet completed.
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1139,9 +1140,15 @@ private void checkDockerIgnore(long startTime, File dockerBuildContext) {
...
             }
-            warn("The docker build command is slower than expected. You may increase performance by adding " +
-                "unneeded files and directories such as any Liberty runtime directories to a .dockerignore file in " +
-                buildContextPath + ".");
+            message += " You may increase performance by adding unneeded files and directories " +
+                "such as any Liberty runtime directories to a .dockerignore file in " +
+                buildContextPath + ".";
+        }
+        if (hasFeaturesSh.get()) {
+            message += " The RUN features.sh command is detected in the Dockerfile and this may increase build time significantly.";
```
ericglau, line 1148: This message sounds as if the user is doing something wrong.  Especially if it started off with "The docker build command is slower than expected."  And since this message appears after the command has finished, should we just avoid printing the message in this case?

turkeylurkey: I thought we should explain why the command is taking so long. I understand it happens every time but it is so slow it bears repeating. If they use cached features hopefully the whole thing finishes in under 30s. Are you saying that if they use RUN features.sh then we should not print the .dockerignore warning/suggestion at all?

turkeylurkey: Sample message with both:
```
[WARNING] The docker build command took longer than 30 seconds. You may increase performance by adding unneeded files and directories such as any Liberty runtime directories to a .dockerignore file in /Users/paulg/WAS/devExperience/demo-devmode. The RUN features.sh command is detected in the Dockerfile and extra time may be necessary for installing features.
```

turkeylurkey:
Sample message appears before docker build:
```
[INFO] Building Docker image...
[INFO] docker build --pull -f /var/folders/p1/v7svjg7s6xngg377xtsjxcmc0000gn/T/tempDockerfile1934724329202521983 -t demo-devmode-maven-----dev-mode /Users/paulg/WAS/devExperience/demo-devmode
[WARNING] The RUN features.sh command is detected in the Dockerfile and extra time may be necessary when installing features.
[INFO] Sending build context to Docker daemon  14.08MB
[INFO] 
[INFO] Step 1/7 : from openliberty/open-liberty
...
```

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1100,6 +1100,9 @@ private void buildDockerImage(File tempDockerfile, File userDockerfile, boolean
             sb.append("-f " + tempDockerfile + " -t " + imageName + " " + userDockerfile.getParent());
             String buildCmd = sb.toString();
             info(buildCmd);
+            if (hasFeaturesSh.get()) {
+                warn("The RUN features.sh command is detected in the Dockerfile and extra time may be necessary when installing features.");
```
ericglau, line 1104: I think we should change this to log as `info` to be less alarming, and since we don't know if installing features will actually take a long time.

turkeylurkey: Ok!

ericglau (approved):

PR #246, opened 2021-01-15
Do not add current classloader to compile classpath

ericglau:
When recompiling in dev mode for Maven, do not include the current classloader (of Liberty Maven plugin) for determining the compile classpath.  Instead, it should only use the artifacts defined in the Maven project for the classpath.

Fixes https://github.com/OpenLiberty/ci.maven/issues/1081


turkeylurkey (approved):
It looks like this seemed to be a good idea at the time.

PR #252, opened 2021-03-31
Add directories to compile classpath

ericglau:
For Maven multi module projects, dev mode adds jars to the classpath and the current project's target classes dir, but should also include other modules' target classes directories as well.


turkeylurkey (approved):
I see you did some clean up, thank you.

PR #262, opened 2021-05-14
Multi-module support for hot source compilation

kathrynkodama:
Adds support for hot compilation of multi module projects. Tracks upstream projects and supports the following scenarios:
- Source compilation of upstream projects
   - change in source code triggers hot compilation of that class. If successful, tries compiling any projects with compilation failures
   - on first watch loop, skips attempting to recompile projects with compilation failures (to avoid multiple compilations if a user starts dev mode with some compile errors)
   - in the multi-module scenario messages will specify the module they relate to, ie. `guide-maven-multimodules-jar source compilation was successful.`

- Watching `pom.xml` of upstream projects
   - updates the classpath used for compilation for the corresponding project
   - tries recompiling any projects with compilation failures (in case adding a dependency to the `pom.xml` solved some compilation errors)

- Watching resource directories of upstream projects
   - supports modifying/deleting resource file, and copying it to the corresponding `target/classes` directory 
   - supports adding a resource directory to an upstream project after dev mode has been started

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -80,7 +80,6 @@ import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.input.CloseShieldInputStream;
 import org.apache.commons.io.monitor.FileAlterationListenerAdaptor;
 import org.apache.commons.io.monitor.FileAlterationObserver;
-import org.apache.maven.artifact.DependencyResolutionRequiredException;
 import org.apache.maven.artifact.versioning.ComparableVersion;
 import org.apache.maven.model.Build;
 import org.apache.maven.project.MavenProject;
@@ -208,6 +207,16 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
      */
     public abstract boolean updateArtifactPaths(File buildFile, List<String> compileArtifactPaths,
             ThreadPoolExecutor executor) throws PluginExecutionException;
+
+    /**
+     * Gets a list of resource directories for the specified project
+     * 
+     * @param project                MavenProject
+     * @param outputDir output directory of corresponding project
+     * @return List of files, if no resource directories are detected, returns a
+     *         list with the only the default resource dir
+     */
+    public abstract List<File> getResourceDirectories(MavenProject project, File outputDir);
 
     /**
      * Run the unit tests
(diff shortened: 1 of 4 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2484,8 +2549,17 @@ public void watchFiles(File buildFile, File outputDirectory, File testOutputDire
                         }
                     }
                 }
+                if (!upstreamProjects.isEmpty()) { // process java compilation for upstream projects
```
ericglau, line 2552: Should this check for null? e.g. `if (upstreamProjects != null && !upstreamProjects.isEmpty()) {`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2561,6 +2635,21 @@ public void watchFiles(File buildFile, File outputDirectory, File testOutputDire
...
+                // check if resourceDirectory of an upstream project has been added
+                for (UpstreamProject p : upstreamProjects) {
+                    for (File resourceDir : p.getResourceDirs()) {
+                        if (!p.getResourceMap().get(resourceDir) && resourceDir.exists()) {
+                            registerAll(resourceDir.getCanonicalFile().toPath(), executor);
+                            p.getResourceMap().put(resourceDir, true);
+                        } else if (p.getResourceMap().get(resourceDir) && !resourceDir.exists()) {
+                            // deleted resource directory
+                            warn("The resource directory " + resourceDir
+                                    + " was deleted.  Restart liberty:dev mode for it to take effect.");
```
ericglau, line 2647: In the message, just say "dev mode" instead of "liberty:dev mode".  Same for the section above (line 2633)

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2965,6 +3139,60 @@ private void processFileChanges(
         // reset this property in case it had been set to true
         System.setProperty(SKIP_BETA_INSTALL_WARNING, Boolean.FALSE.toString());
 
+        // upstream project source file changed
+        if (!this.upstreamProjects.isEmpty()) {
```
ericglau, line 3143: Should this check for null? e.g. `if (upstreamProjects != null && !upstreamProjects.isEmpty()) {`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2911,6 +3076,15 @@ private void initWatchLoop() throws IOException {
         triggerJavaSourceRecompile = false;
         triggerJavaTestRecompile = false;
 
+        // initial source compile of upstream projects
+        for (UpstreamProject project : this.upstreamProjects) {
```
ericglau, line 3080: Is the `this.` prefix necessary?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -3652,14 +3886,14 @@ protected boolean recompileJava(Collection<File> javaFilesChanged, List<String>
             }
             if (compileResult) {
                 if (tests) {
-                    info("Tests compilation was successful.");
+                    info(formatMessage("Tests compilation was successful", projectName));
```
ericglau, line 3889: Generally we should avoid using methods to format messages, so that it will be easier to switch to localized messages later on.  Use separate message strings for the different situations.  Same for the other messages in this method.

ericglau (approved):

PR #263, opened 2021-06-06
Support hot compilation of and running tests for multi module Maven projects

kathrynkodama:
Adds support for hot compilation of test directories for multi module Maven projects, and supports running tests across all modules.
- Dev mode watches the test directories of the different modules, and on save will recompile. On a successful recompile, dev mode will try to recompile any other failing classes within the same module. 
- On Enter, all tests across all modules will be run. The `upstreamProjects` object is now a List instead of a Set, to ensure that when tests are run for all modules, they will match the reactor build order. 
- If the `hotTests` flag is enabled, only the tests corresponding to the module changed will be run (the `runTests` method now accepts build file(s), and will run tests against the build file(s) specified). If the `hotTests` flag is enabled, handle the initial run of tests after java compilation in the `processJavaCompilation` method. 
- Honour the values for `skipTests`, `skipUTs` or `skipITs` that are configured via the pom.xml in any of the upstream modules. Any values specified via command line (`-DskipITs`) will take precedence. As with regular dev mode, unit tests will always be skipped for `ear` Maven modules.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -295,6 +295,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
     private Path tempConfigPath;
     private boolean skipTests;
     private boolean skipUTs;
+    private boolean forceSkipUTs;
     private boolean skipITs;
     private String applicationId;
     private int appStartupTimeout;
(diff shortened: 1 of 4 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4152,5 +4351,34 @@ public String getContainerName() {
...
+     * Checks if upstreamProjects exist, indicating that this is a multi module
+     * project
+     * 
+     * @return true if this is a multi module project, false if not
+     */
+    private boolean isMultiModuleProject() {
+        if (upstreamProjects != null && !upstreamProjects.isEmpty()) {
+            return true;
+        }
+        return false;
```
ericglau, line 4364: Could simplify with 
```
return (upstreamProjects != null && !upstreamProjects.isEmpty());
```

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4021,23 +4185,58 @@ public void runTestThread(boolean waitForApplicationUpdate, ThreadPoolExecutor e
...
+                            // the skipUTs value
+                            for (UpstreamProject upstreamProject : upstreamProjects) {
+                                if (upstreamProject.getBuildFile().equals(currentBuildFile)) {
+                                    currentProjForceSkipUTs = upstreamProject.skipUTs();
+                                    currentProjForceSkipITs = upstreamProject.skipITs();
+                                    currentProjSkipTests = upstreamProject.skipTests();
+                                    currentProjName = upstreamProject.getProjectName();
+                                    break;
+                                }
+                            }
```
ericglau, line 4222: This is creative :)

ericglau (approved):

PR #266, opened 2021-06-18
Add env var to allow remote debug

ericglau:
When devc issues the `docker run` command, append [WLP_DEBUG_REMOTE environment var](https://github.com/OpenLiberty/open-liberty/blob/2fd4a880754c37a988c5ed9ac4f1ea5988e465d6/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server#L58) as WLP_DEBUG_REMOTE=y to allow remote debugging on newer Java versions such as 11.

Fixes https://github.com/OpenLiberty/ci.maven/issues/1161


turkeylurkey (approved):

scottkurz (approved):
LGTM

I wonder if there's just as much of a case for just adding this in dev mode in all cases beyond just devc?

I suppose that could be done later..and people can add it themselves.

ericglau:
For regular dev mode, unless there is a scenario where remote debugging is common, users should enable it themselves if needed. Otherwise it could be a security concern.

PR #272, opened 2021-07-07
Add recompileDependencies option for dev mode and default behaviour

kathrynkodama:
**Java compilation**
Part of https://github.com/OpenLiberty/ci.maven/issues/1174

Default behaviour for `recompileDependencies` is set in ci.maven/ci.gradle.
When `recompileDependencies=true` recompile all classes in the current and dependent modules when a:
- src/main/java file is added, modified or deleted
- src/test/java files is added, modified or deleted
- build file changes

For multi module projects, when `recompileDependencies=true` always compile in build order: 
1) Failing classes from any upstream modules
2) All classes in the current module
3) All classes in any dependent (downstream) modules

DevUtil uses a `disableDependencyCompile` flag for each project to only compile the current module. This is needed to avoid duplication compilation is step (3) above. 

**multi module hotTests**
Part of https://github.com/OpenLiberty/ci.maven/issues/1175

Run tests on current module and dependent (downstream) modules when:
- a resource directory has been added or deleted
- resource files have been added, modified or deleted
- Java files were deleted without any other changes

To complete the `hotTests` story for multimodules we still need to implement running tests on all dependent modules.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2411,7 +2411,8 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
     WatchService watcher;
 
     // used for multi module projects
-    Set<ProjectModule> upstreamProjectsWithCompilationErrors;
+    Set<ProjectModule> projectsWithSourceCompilationErrors;
+    Set<ProjectModule> projectsWithTestCompilationErrors;
     boolean triggerUpstreamJavaSourceRecompile;
     boolean initialCompile;
     boolean disableDependencyCompile;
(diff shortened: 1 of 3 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2965,21 +2976,30 @@ private void processUpstreamJavaCompilation(List<UpstreamProject> upstreamProjec
                     if (initialCompile) {
                         skipRunningTests = true;
                     }
+
+                    // try recompiling failing project modules that are not dependent on the current
+                    // module (upstream of the current module)
+                    if (!project.disableDependencyCompile && recompileDependencies && !initialCompile) {
```
ericglau, line 2982: Since this combination of 3 booleans is used several times throughout this class, should it be combined into a boolean method for convenience?  e.g. something like `shouldRecompileDependencies(ProjectModule project)` or some other appropriate name

ericglau (approved):

PR #275, opened 2021-07-14
Add hotTests logic for multi module projects

kathrynkodama:
Partial fix for https://github.com/OpenLiberty/ci.maven/issues/1175 (integration tests in ci.maven still need to be added) 

When `hotTests=true` for multi-module projects, always run tests on the modified module and all dependent modules.

Since we have to run tests using the same `ThreadPoolExecutor`, for multi-module projects do not run tests through `recompileJavaSource` or `recompileJavaTest` (that is why [skipRunningTests is true](https://github.com/OpenLiberty/ci.common/pull/275/files#diff-d8b4217f048364ad127bea282a1726d1523aa8359775cee2f4b71eec9cba7140R2979-R2985) when called from `processUpstreamJavaCompilation`). Instead, call `runTestThread()` directly and pass in the current build file and all dependent build files. 

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2975,7 +2975,6 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                 }
             }
             for (ProjectModule project : upstreamProjects) {
-
                 // delete before recompiling, so if a file is in both lists, its class
                 // will be deleted then recompiled
                 if (!project.deleteJavaSources.isEmpty()) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2999,8 +2994,12 @@ private void processUpstreamJavaCompilation(List<ProjectModule> upstreamProjects
                         for (File dependentModule : project.getDependentModules()) {
                             compileModuleForBuildFile(dependentModule, false);
                         }
-                        project.disableDependencyCompile = true;
                     }
+                    // run tests for current project and all dependent projects
+                    int numApplicationUpdatedMessages = countApplicationUpdatedMessages();
+                    runTestThread(true, executor, numApplicationUpdatedMessages, skipUTs, false,
```
ericglau, line 3000: Should this be project.skipUTs()?

ericglau (approved):

PR #285, opened 2021-07-27
Multi-module projects run tests when all compilations are successful

kathrynkodama:
For multi-module projects with `-DhotTests=true`, only run tests when all compilations are successful. 

For multi module projects with `-DrecompileDependencies=true` (default) and `-DhotTests=true`, compilation occurs in the following order:
1. Compile any failing source and test classes in upstream modules 
2. Compile all source and test classes in the module where the file was changed
3. Compile all source and test classes in the dependent modules
4. if ALL of the above compilations were successful, run tests on the current module (step 2) and all dependent modules (step 3)

*if a test file was changed, steps 1-4 are the same except only test classes will be compiled

For multi module projects with `-DrecompileDependencies=false` and `-DhotTests=true`, compilation occurs in the following order:
1. Compile any failing source and test classes in upstream modules 
2. Compile the file changed
3. Compile any failing source and test classes in dependent modules
4. if ALL of the above compilations were successful, run tests on the current module (step 2) and all dependent modules (step 3)

*if a test file was changed, steps 1-4 are the same except only test classes will be compiled

Integration tests for the above behaviour will be added in a follow-up PR in ci.maven. 

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -4691,39 +4691,11 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
     }
 
     /**
-     * Compile all classes in the module corresponding to the build file
-     * 
-     * @param moduleBuildFile build file
-     * @param testsOnly       true if only test classes should be compiled
-     * @param executor        ThreadPoolExecutor
-     * @return true if compilation is successful
-     * @throws IOException
-     * @throws PluginExecutionException
-     */
-    private boolean compileModuleForBuildFile(File moduleBuildFile, boolean testsOnly, ThreadPoolExecutor executor)
-            throws IOException, PluginExecutionException {
-        if (moduleBuildFile.getCanonicalPath().equals(buildFile.getCanonicalPath())) {
-            debug("recompileDependencies is set to true, recompiling the entire module for "
-                    + moduleBuildFile.getCanonicalPath());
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4696,7 +4762,61 @@ private void compileEntireProject(File sourceDir, Collection<File> recompileJava
...
+    private boolean compileAllClasses(ProjectModule project, boolean tests, ThreadPoolExecutor executor)
+            throws IOException, PluginExecutionException {
+        return compileAllClasses(project.getSourceDirectory(), project.getTestSourceDirectory(),
+                project.getPackagingType(), project.getCompileArtifacts(), project.getTestArtifacts(),
+                project.getOutputDirectory(), project.getTestOutputDirectory(), project.getProjectName(),
+                project.getBuildFile(), project.getCompilerOptions(), project.skipUTs(),
+                project.failedCompilationJavaSources, project.failedCompilationJavaTests, tests, executor);
+    }
+
+    private boolean compileAllClasses(File sourceDir, File testSourceDir, String packagingType,
```
ericglau, line 4782: Can you add some javadoc comments to clarify the differences between these methods and `triggerMainModuleCompile`/`triggerUpstreamModuleCompile`?  So that it the distinction is clearer in case future maintenance is needed.

ericglau (approved):
Looks good.  Just one suggestion.

PR #287, opened 2021-08-04
Initialize empty parent build file map if null

kathrynkodama:
Part of https://github.com/OpenLiberty/ci.maven/pull/1228 


ericglau (approved):
It would be useful to also add a javadoc comment above line 379 to describe the mappings represented in the variable `parentBuildFiles`

PR #292, opened 2021-09-20
Read config document from file and process as server.xml with featureManager element

turkeylurkey:
This code is in ServerConfigDropinXmlDocument which is not a great name for the purpose to which it is now being put. This class is now used also to handle server.xml files where previously it just created the variable xml files and the generated features xml files.

We could create a new class with a name such as ServerConfigXmlDocument which extends ServerConfigDropinXmlDocument if we want to separate the server.xml handling. I'm not thrilled with that hierarchy so we could create a new common sub-class both these classes could extend.

Let me know what you think.

Also, do you know why XmlDocument wants to ignore comments?

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDropinXmlDocument.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDropinXmlDocument.java
@@ -15,22 +15,40 @@
  */
 package io.openliberty.tools.common.plugins.config;
 
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
 import javax.xml.parsers.ParserConfigurationException;
 
 import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+import org.w3c.dom.Text;
+import org.xml.sax.SAXException;
 import org.w3c.dom.Element;
 import org.w3c.dom.Comment;
 
 public class ServerConfigDropinXmlDocument extends XmlDocument {
-    
+
     private Element featureManager = null;
 
-    private ServerConfigDropinXmlDocument() {    
+    private ServerConfigDropinXmlDocument() {
     }
-    
+
     public static ServerConfigDropinXmlDocument newInstance() throws ParserConfigurationException {
         ServerConfigDropinXmlDocument configDocument = new ServerConfigDropinXmlDocument();
         configDocument.createDocument("server");
+        return configDocument;
+    }
+
+    public static ServerConfigDropinXmlDocument newInstance(File f) throws ParserConfigurationException, SAXException, IOException {
+        if (f == null || !f.exists()) {
+            return null;
+        }
+        ServerConfigDropinXmlDocument configDocument = new ServerConfigDropinXmlDocument();
+        configDocument.createDocument(f);
+        configDocument.featureManager = configDocument.findFeatureManager();
         return configDocument;
     }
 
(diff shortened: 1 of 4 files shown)


cherylking:
@turkeylurkey I noticed you looked like you are having difficulties with the format of the written xml file. So I just wanted to drop this code fragment here that I use elsewhere to write a xml file. It might be worth a try.

```
    public static void writeDocToXmlFile(Document doc, File inputFile) throws Exception {
        TransformerFactory transformerFactory = TransformerFactory.newInstance();
        Transformer transformer = transformerFactory.newTransformer();
        transformer.setOutputProperty(OutputKeys.METHOD, "xml");
        transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-8");
        transformer.setOutputProperty(OutputKeys.INDENT, "yes");
        transformer.setOutputProperty(OutputKeys.STANDALONE, "yes");
        transformer.setOutputProperty("{http://xml.apache.org/xslt}indent-amount", "4");
        transformer.setOutputProperty(OutputKeys.DOCTYPE_PUBLIC, "yes");

        doc.setXmlStandalone(true);
        DOMSource source = new DOMSource(doc);
        StreamResult file = new StreamResult(new OutputStreamWriter(new FileOutputStream(inputFile), "UTF-8"));
        transformer.transform(source, file);
  }
```

turkeylurkey:
@cherylking Thank you, I'll compare that to the code we have for XML handling. My main issue was the difference between when you create a DOM from scratch and when it is created by parsing an XML file. The object model is different. There was some learning curve there.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDropinXmlDocument.java
```diff
@@ -15,38 +15,109 @@
...
     }
-    
+
     public static ServerConfigDropinXmlDocument newInstance() throws ParserConfigurationException {
         ServerConfigDropinXmlDocument configDocument = new ServerConfigDropinXmlDocument();
         configDocument.createDocument("server");
         return configDocument;
     }
 
+    public static ServerConfigDropinXmlDocument newInstance(File f) throws ParserConfigurationException, SAXException, IOException {
```
ericglau, line 45: It would be useful to add javadoc comments to describe this method e.g. what is `File f` parameter for

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDropinXmlDocument.java
```diff
@@ -68,4 +139,47 @@ public void createFeature(String name) {
         child.appendChild(text);
         featureManager.appendChild(child);
     }
+
+    public Element findFeatureManager() {
+        Element serverElement = findServerElement();
+        if (serverElement == null) {
+            return null;
+        }
+        Node fmNode = serverElement.getElementsByTagName("featureManager").item(0); // assume just one
```
ericglau, line 148: Can `getElementsByTagName("featureManager")` return null if the `featureManager` element is not found?  If so, should check for a null pointer before calling `.item(0);`

turkeylurkey: `getElementsByTagName("featureManager")` will always return a NodeList which can be empty. `item()` will return a null if the index is out of bounds and not throw an exception like IndexOutOfBounds so this is ok. `fmNode` could be null in which case the `instanceof` fails.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDropinXmlDocument.java
turkeylurkey: 👍

PR #297, opened 2021-10-05
Enhance class to return the list of server features with original mixed case

turkeylurkey:
under control of an option.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
@@ -122,7 +122,7 @@ public abstract class ServerFeatureUtil extends AbstractContainerSupportUtil {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -120,6 +121,15 @@
         return getConfigDropinsFeatures(result, serverDirectory, bootstrapProperties, "overrides");
     }
 
+    /**
+     * Indicate that the feature names should remain unmodified. The default is to make all the names lower case.
```
ericglau, line 125: `Indicate that the feature names should remain unmodified.` -> `Indicate whether the feature names should be converted to lowercase.`

ericglau (approved):

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
turkeylurkey: 👍

PR #301, opened 2021-10-18
Add check for valid classpath, block tests from running if invalid

kathrynkodama:
See https://github.com/OpenLiberty/ci.maven/pull/1260

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -5007,13 +5007,13 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
 
     // prints error to users when Maven thread classpath error is present
     private void printMavenClasspathErrMsg(boolean startup) {
+        String message1 = "A classpath error with Maven has been detected. Tests will not run in dev mode on this project.";
+        String message2 = "Run tests outside of dev mode with `mvn verify` or `mvn failsafe:integration-test`.";
         if (startup) {
-            String message1 = "A classpath error with Maven has been detected. Tests will not run in dev mode on this project.";
-            String message2 = "Run tests outside of dev mode with `mvn verify` or `mvn failsafe:integration-test`.";
-            error(startup ? formatAttentionMessage(message1) : message1);
-            error(startup ? formatAttentionMessage(message2) : message2);
+            error(formatAttentionMessage(message1));
+            error(formatAttentionMessage(message2));
         } else {
-            error("A classpath error with Maven has been detected. Tests will not run in dev mode on this project. Run tests outside of dev mode with `mvn verify` or `mvn failsafe:integration-test`.");
+            error(message1 + " " + message2);
         }
     }
 }


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2238,12 +2254,19 @@ private void printDevModeMessages(boolean inputUnavailable, boolean startup) {
...
         if (!inputUnavailable) {
             // the following will be printed on startup and every time after the tests run
-            if (hotTests) {
-                String message = "Tests will run automatically when changes are detected. You can also press the Enter key to run tests on demand.";
-                info(startup ? formatAttentionMessage(message) : message);
+            if (blockTests) {
+                String message1 = "A classpath error with Maven has been detected. Tests will not run in dev mode on this project.";
+                String message2 = "Run tests outside of dev mode with `mvn verify` or `mvn failsafe:integration-test`.";
+                info(startup ? formatAttentionMessage(message1) : message1);
+                info(startup ? formatAttentionMessage(message2) : message2);
```
ericglau, line 2261: Suggest to log these as warning or error so that they are more visible, especially since they are part of the dev mode startup messages and users may not notice those otherwise.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -3396,13 +3427,18 @@ private void processJavaCompilation(File outputDirectory, File testOutputDirecto
             if (initialCompile) {
                 initialCompile = false;
                 if (hotTests) {
-                    // if hot testing, run tests on startup
-                    if (isMultiModuleProject()) {
-                        // force run tests across all modules in multi module scenario
-                        runTestThread(false, executor, -1, true, getAllBuildFiles());
+                    if (blockTests) {
+                        error("A classpath error with Maven has been detected. Tests will not run in dev mode on this project. Run tests outside of dev mode with `mvn verify` or `mvn failsafe:integration-test`.");
```
ericglau, line 3431: Can this just call `printMavenClasspathErrMsg()`?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: We've done something similar with devc [here](https://github.com/OpenLiberty/ci.common/blob/main/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java#L2277) for port warnings, see example:
Screen Shot 2021-10-20 at 10 56 50 AM
I think the formatting being off is fine.  At least having something different there draws attention.

kathrynkodama: Updated to print an error message and removed the error message that displays immediately after the startup message.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -5012,7 +5006,14 @@ private boolean isUpstreamSourceDir(File dirAdded) {
...
 
     // prints error to users when Maven thread classpath error is present
-    private void printMavenClasspathErrMsg() {
-        error("A classpath error with Maven has been detected. Tests will not run in dev mode on this project. Run tests outside of dev mode with `mvn verify` or `mvn failsafe:integration-test`.");
+    private void printMavenClasspathErrMsg(boolean startup) {
+        if (startup) {
+            String message1 = "A classpath error with Maven has been detected. Tests will not run in dev mode on this project.";
+            String message2 = "Run tests outside of dev mode with `mvn verify` or `mvn failsafe:integration-test`.";
+            error(startup ? formatAttentionMessage(message1) : message1);
+            error(startup ? formatAttentionMessage(message2) : message2);
```
ericglau, line 5014: In this code block within the `if` condition, `startup` is already `true` so only the `formatAttentionMessage()` part is being used here.

PR #307, opened 2021-11-04
Track class files for feature generation

kathrynkodama:
- track source class files for feature generation
- on class file change, pass changed class files to the Liberty feature generation

See https://github.com/OpenLiberty/ci.maven/pull/1275
See https://github.com/OpenLiberty/ci.gradle/pull/640

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2700,6 +2700,17 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                             testArtifactPaths, null, false);
                 }
 
+                // when change in class files are detected scan for Liberty features.
+                if (generateFeatures && !initialCompile && !javaSourceClasses.isEmpty()) {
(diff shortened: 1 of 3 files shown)


ericglau (approved):

turkeylurkey (commented):
There is code in processJavaCompilation() so that when we compile Java we call libertyGenerateFeatures(). Can we remove this code?

kathrynkodama:
> There is code in processJavaCompilation() so that when we compile Java we call libertyGenerateFeatures(). Can we remove this code?

This call to libertyGenerateFeatures() left in processJavaCompilation() is only for initial compile of Gradle projects. This is because when dev mode starts, it does its own initial compilation of all the java files to push any compilation errors to the top of the console. For Maven, when we compile the java files we regenerate all of the source class files and they are added to our tracking list.

For Gradle however, the Gradle compiler will only recompile classes that have changed since it was last called, so the source class files may not be regenerated and therefore not added to our tracking list. Then for our initial Gradle compile we have to leave in the call for libertyGenerateFeatures() with empty params so that it scans every class file.

turkeylurkey (approved):

TrevCraw (approved):

PR #309, opened 2021-11-17
Add webResource watching

awisniew90:
Signed-off-by: Adam Wisniewski <awisniew@us.ibm.com>

Add web resource checking for ci.maven "exploded" loose war app support

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -320,6 +320,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
     private File projectDirectory;
     private File multiModuleProjectDirectory;
     private List<File> resourceDirs;
+    private List<Path> webResourceDirs;
     private boolean hotTests;
     private Path tempConfigPath;
     private boolean skipTests;
@@ -400,7 +401,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
             boolean skipDefaultPorts, JavaCompilerOptions compilerOptions, boolean keepTempDockerfile,
             String mavenCacheLocation, List<ProjectModule> upstreamProjects, boolean recompileDependencies,
             String packagingType, File buildFile, Map<String, List<String>> parentBuildFiles,
-            Set<String> compileArtifactPaths, Set<String> testArtifactPaths) {
+            Set<String> compileArtifactPaths, Set<String> testArtifactPaths, List<Path> webResourceDirs) {
         this.buildDirectory = buildDirectory;
         this.serverDirectory = serverDirectory;
         this.sourceDirectory = sourceDirectory;
@@ -466,6 +467,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
         this.compileArtifactPaths = compileArtifactPaths;
         this.testArtifactPaths = testArtifactPaths;
         this.blockTests = false;
+        this.webResourceDirs = webResourceDirs;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2456,6 +2469,7 @@ private void readInput() {
     long lastJavaSourceChange;
     long lastJavaTestChange;
     Map<File, Long> lastBuildFileChange;
+    boolean triggerAppRefresh;
```
scottkurz, line 2472: I think maybe this was deleted since mine but you added it back in?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4429,6 +4500,8 @@ protected boolean recompileJava(Collection<File> javaFilesChanged, Set<String> a
                     // redeploy app after compilation if not loose application
                     if (!isLooseApplication()) {
                         redeployApp();
+                    } else {
+                	    updateLooseApp();
```
scottkurz, line 4504: indentation is off.. easy to fix?

scottkurz (approved):
Looks good in and of itself though I didn't get all the way through the ci.maven change

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -3877,6 +3936,18 @@ private ProjectModule getFirstProjectModule(ProjectModule project, File buildFil
...
+
+	}
+
+	protected void resourceModifiedOrCreated(File fileChanged, File resourceParent, File outputDirectory) throws IOException {
+        copyFile(fileChanged, resourceParent, outputDirectory, null);
+	}
+
+    protected void resourceDeleted(File fileChanged, File resourceParent, File outputDirectory) throws IOException {
+        deleteFile(fileChanged, resourceParent, outputDirectory, null);
+	}
```
ericglau, line 3950: Indentation is off here as well

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1132,6 +1134,17 @@ protected void processCopyLines(List<String> dockerfileLines, String buildContex
...
+    
+    /**
+     * 
+     * The intention here is to give a chance to run logic for an app update
+     * but not necessarily a full re-deploy
+     * 
+     * @throws PluginExecutionException
+     */
+    protected void updateLooseApp() throws PluginExecutionException {
+    	// no-op - placeholder to override
```
ericglau, line 1146: If this is meant to be overridden in ci.maven, can you change this to an `abstract` method?  (This will mean it needs to be overridden in ci.gradle as well, but that can be a no-op if it's not applicable for Gradle).

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
ericglau: So these methods are overriden in ci.maven, meaning this default implementation is only used with cigradle.  While there is nothing wrong with this, I'd prefer to make these `abstract` and override/implement them separately in ci.maven and ci.gradle to avoid confusion.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
awisniew90: Good catch.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
awisniew90: Done

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
awisniew90: Done

ericglau (approved):

PR #311, opened 2021-11-18
Do not watch files that are in the source directory and the loose app config file

kathrynkodama:
Adds logic to not watch files that are in the source directory and specified in the loose application file.
Part of https://github.com/OpenLiberty/ci.maven/issues/1285

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2564,7 +2564,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                     if (shouldIncludeSources(p.getPackagingType())) {
                         // watch src/main/java dir
                         if (p.getSourceDirectory().exists()) {
-                            omitWatchingFiles.addAll(getOmitFilesList(looseAppFile, p.getSourceDirectory().getCanonicalFile().toPath()));
+                            omitWatchingFiles.addAll(getOmitFilesList(looseAppFile, p.getSourceDirectory().getCanonicalPath()));
                             registerAll(p.getSourceDirectory().getCanonicalFile().toPath(), executor);
                             p.sourceDirRegistered = true;
                         }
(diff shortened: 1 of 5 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -3050,6 +3069,50 @@ private void onAlteration(final ThreadPoolExecutor executor, final String parent
...
+                            File srcOnDiskFile = new File(srcOnDiskNodeText);
+                            if (srcOnDiskFile.getCanonicalPath().startsWith(srcDirectoryPath)) {
+                                omitFiles.add(srcOnDiskFile);
+                            }
+                        }
+                    }
+                }
+            }
+        } catch (ParserConfigurationException | SAXException | IOException e) {
+            debug("Unable to read loose application configuration file: " + looseAppFile.toString());
```
ericglau, line 3110: Should this log as error so that any underlying issues would be noticed?  Although there is not much the user can do in this case.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4177,7 +4243,12 @@ public FileVisitResult preVisitDirectory(final Path dir, BasicFileAttributes att
                                 return FileVisitResult.CONTINUE;
                             }
                         }
-        
+                        for (File omitFile : omitWatchingFiles) {
+                            if (dir.startsWith(omitFile.getCanonicalPath())) {
```
ericglau, line 4247: I realized a number of places in DevUtil do something like this, but take for example:
File A has path "/abc/somepath"
File B has path "/abc/somepathwithsubstring/something"
Then `B.startsWith(A.getCanonicalPath())` would return true, even though B is not a child of A.  

The solution could be to check for A's path along with File.separator, or check B.getParent() recursively, but that could be a separate issue since other places in DevUtil have the same issue.  And this is very much a corner case.

ericglau (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
kathrynkodama: I've logged this as an error so at least users will be aware

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
kathrynkodama: I've added the check for A's path along with File.separator for this code. As for the rest of DevUtil, we can open a follow up issue if you think it is would be an improvement

ericglau: I'm not sure if File.separator would cause any issues between OS'es.  One scenario that comes to mind is running devc on Windows (but with container's base image using Linux) -- the loose app file is created locally but mounted for the container to use.  I guess this would be fine since dev mode is running locally, but something to test out.

ericglau (approved):

PR #313, opened 2021-11-26
Write XML document without indentation if the document has indentation embedded in it.

turkeylurkey:
Also return true if adding a comment to XML document is successful. Update test.

The document will have indentation embedded in the case it is read from a file by the parser. The embedded spaces appear as Text nodes which contain, for example, \n<blank><blank><blank><blank>. The code currently checks for these spaces. 

Another design would be to overload the setXmlStandalone() flag which is set when the document is created internally and is currently not set by the parser when the document is read from a file. We could control the indentation using that flag. 

Alternatively do the reviewers know of other W3C Document (org.w3c.dom.Document) properties which could indicate embedded indentation whitespace?

Fixes https://github.com/OpenLiberty/ci.maven/issues/1301


ericglau:
This indentation detection seems fine.  We can look for alternatives if we notice issues during further testing.

ericglau (approved):

PR #320, opened 2021-12-02
Generate features from dev mode

TrevCraw:


src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -167,17 +167,18 @@ private ClassLoader getScannerClassLoader() throws MalformedURLException {
...
+        if (!optimize && classFiles != null && !classFiles.isEmpty()) {
             resultSet = classFiles;
-        } else {
+        } else if (optimize) {
             if (classDirectories == null || classDirectories.isEmpty()) {
                 throw new PluginExecutionException("Error collecting list of directories to send to binary scanner, list is null or empty.");
             }
             resultSet = classDirectories;
+        } else {
+            return new String[0];
```
ericglau, line 180: This set of if-statements is confusing because they have multiple conditions and it's not immediately obvious what are the conditions for the last `else` .  Seems that is for !optimize and classFiles are null or empty.  In this case, consider rearranging the conditions to the following for better readability:
```
if (optimize) {
  ...
} else {
  if (classFiles != null && !classFiles.isEmpty())) {
    ...
  } else {
    ...
  }
}

**Review by ericglau — APPROVED** [review]

File: src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java [hunk 2]
```diff
@@ -2422,22 +2423,38 @@ private void toggleFeatureGeneration() {
...
+    private void incrementGenerateFeatures() {
+        info("Generating feature list from incremental changes...");
+        try {
+            Collection <String> javaSourceClassPaths = getClassPaths(javaSourceClasses);
+            libertyGenerateFeatures(javaSourceClassPaths, false);
+        } catch (PluginExecutionException | IOException e) {
+            error("An error occurred while trying to generate features: " + e.getMessage(), e);
+        }
+        // TODO: update clear logic to handle cases where feature generation fails
+        javaSourceClasses.clear();
```
kathrynkodama, line 2455: It looks like if feature generation fails `javaSourceClasses` will still be cleared. Wondering if this line should be after 2450:
```java
            libertyGenerateFeatures(javaSourceClassPaths, false);
            javaSourceClasses.clear();
        } catch (PluginExecutionException | IOException e) {
```

kathrynkodama: Maybe on error feature generation should be turned off?

TrevCraw: Both of these suggestions are good. I plan to address dealing with the error/exception scenarios in another PR. I'm still not clear on the flow for the feature generation goal failing (ie. when we fail it on our terms vs something else goes wrong). I was thinking of catching that earlier on and maybe having a return value for `libertyGenerateFeatures()`. So if an Exception occurs here, it will be one that was not expected to be caught. So the second suggestion may work well here.

kathrynkodama (approved):

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
TrevCraw: I will rearrange the logic here.

kathrynkodama (approved):

PR #325, opened 2021-12-13
Generate features

TrevCraw:
This merge is to complete part of automatic generation of features in server configuration: https://github.com/OpenLiberty/open-liberty/issues/12983


turkeylurkey (approved):

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -0,0 +1,263 @@
...
+
+    // The jar file containing the binary scanner code
+    private File binaryScanner;
+    private URLClassLoader binaryScannerClassLoader = null;
+    
+    public BinaryScannerUtil(File binaryScanner) {
+        this.binaryScanner = binaryScanner;
+    }
+
+    public Set<String> runBinaryScanner(Set<String> currentFeatureSet, List<String> classFiles, Set<String> allClassesDirectories,
```
cherylking, line 58: It would be good to guard against an infinite loop since this method is called recursively. There should be a condition that could be checked to bail out at the beginning of the method if the calls have gone more levels than expected. You can open a separate issue to address after the snapshot.

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -0,0 +1,263 @@
...
+                    warn("Caused by exception:"+x.getCause().getClass().getName());
+                    warn("Caused by exception message:"+x.getCause().getMessage());
+                }
+                error(x.getMessage());
+            }
+        } else {
+            if (binaryScanner == null) {
+                throw new PluginExecutionException("The binary scanner jar location is not defined.");
+            } else {
+                throw new PluginExecutionException("Could not find the binary scanner jar at " + binaryScanner.getAbsolutePath());
```
cherylking, line 149: As we discussed on the call, I don't think we should throw an exception when the binary scanner is not found. Instead log a warning that the features could not be generated because the binary scanner could not be located.

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -1009,12 +1022,27 @@ private void installFeaturesOnContainer(List<String> features, boolean acceptLic
...
         } else {
             // Log the successful output as debug
             debug(cmdResult);
         }
     }
-    
+
+    // Return true if feature conflict code is detected in the exception message
+    private boolean isFeatureConflict(String exceptionMessage) {
+        Pattern p = Pattern.compile(ANY_CONFLICT);
```
cherylking, line 1040: You can store the Pattern in a static variable since ANY_CONFLICT does not change between invocations.

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -1009,12 +1022,27 @@ private void installFeaturesOnContainer(List<String> features, boolean acceptLic
...
+        Matcher m = p.matcher(exceptionMessage);
+        if (m.find()) {
+            return true;
+        }
+        return false;
+    }
```
cherylking, line 1046: You can simplify these lines to just `return m.find();`

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -102,21 +103,59 @@
...
             libertyDirectoryPropertyToFile = new HashMap<String,File>();
         }
         Properties bootstrapProperties = getBootstrapProperties(new File(serverDirectory, "bootstrap.properties"));
-        Set<String> result = getConfigDropinsFeatures(null, serverDirectory, bootstrapProperties, "defaults");
-        result = getServerXmlFeatures(result, new File(serverDirectory, "server.xml"), bootstrapProperties, null);
+        Set<String> result = getConfigDropinsFeatures(null, serverDirectory, bootstrapProperties, "defaults", dropinsFilesToIgnore);
+        if (serverXmlFile == null) {
+            serverXmlFile = new File(serverDirectory, "server.xml");
+        }
+        result = getServerXmlFeatures(result, serverXmlFile, bootstrapProperties, null);
```
cherylking, line 124: One potential problem to passing in the serverXmlFile is the case where it has `include` elements with relative paths to other config files that are located in their configDir. When the serverXmlFile gets copied to the target server along with the contents of the configDir, everything would resolve correctly. But potentially they specify a serverXmlFile from a different location, and now the includes cannot be found correctly to find all the features. This is related to the calls to `parseIncludeNode` in this class.

cherylking (changes requested):

pom.xml
```diff
@@ -55,6 +55,16 @@
...
+        <dependency>
+            <groupId>javax.json</groupId>
+            <artifactId>javax.json-api</artifactId>
+            <version>1.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.json</artifactId>
+            <version>1.1</version>
+        </dependency>
```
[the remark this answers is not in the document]
turkeylurkey: These were used but no longer. I opened a ci.common issue to track their removal.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigXmlDocument.java
```diff
@@ -0,0 +1,213 @@
+/**
+ * (C) Copyright IBM Corporation 2019, 2021.
```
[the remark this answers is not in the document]
turkeylurkey: Agreed, can we fix it at this point? `git mv` ?

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
kathrynkodama: This will be addressed as part of this issue: https://github.com/OpenLiberty/ci.maven/issues/1291

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigXmlDocument.java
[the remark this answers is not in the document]
cherylking: I'm not enough of a git expert to figure that out. Can you just add a comment in this new file that explains it replaces the other class and gives the path of the original file?

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: Looking at this.

TrevCraw: Issue opened here: https://github.com/OpenLiberty/ci.maven/issues/1341

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigXmlDocument.java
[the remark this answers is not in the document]
turkeylurkey: Ok

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: Recorded as https://github.com/OpenLiberty/ci.maven/issues/1341 and fixed in https://github.com/OpenLiberty/ci.common/pull/328

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
kathrynkodama: I will address this

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
kathrynkodama: I will address this

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
kathrynkodama: Fixed in https://github.com/OpenLiberty/ci.common/pull/330

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
kathrynkodama: Fixed in https://github.com/OpenLiberty/ci.common/pull/330

cherylking (approved):

PR #327, opened 2021-12-14
Improve exception handling for feature generation

kathrynkodama:
Signed-off-by: Kathryn Kodama <kathryn.s.kodama@gmail.com>

PR #328, opened 2021-12-14
Remove unused dependencies and protect recursion in runBinaryScanner().

### LINKED ISSUE #326: Remove JSON dependencies from pom
Two dependencies containing 'javax.json'

turkeylurkey:
Only one more recursion check is needed on top of the new plugin exceptions added in #327 
Fixes #326
Fixes https://github.com/OpenLiberty/ci.maven/issues/1341

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,16 +56,6 @@
             <scope>runtime</scope>
         </dependency>
         <dependency>
-            <groupId>javax.json</groupId>
-            <artifactId>javax.json-api</artifactId>
-            <version>1.0</version>
-        </dependency>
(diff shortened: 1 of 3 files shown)


kathrynkodama (approved):

kathrynkodama (commented):
Per https://github.com/OpenLiberty/ci.maven/issues/1341 should we add an explicit flag to ensure there is only one level of recursion?

turkeylurkey:
We are already using the first argument as a flag because we need to remove the current features passed in when we generate suggestions. I added several comments to make this more explicit.

PR #332, opened 2021-12-17
1332 install order

TrevCraw:
Fixes https://github.com/OpenLiberty/ci.maven/issues/1332

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -4003,6 +4003,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                 boolean generateFeaturesSuccess = false;
                 if (generateFeatures) {
                     // custom server.xml modified
+                    // TODO: revisit scenarios in which we should skip install features
                     generateFeaturesSuccess = incrementGenerateFeatures();
                 }
                 // suppress install feature warning - property must be set before calling copyConfigFolder
@@ -4042,6 +4043,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                 boolean generateFeaturesSuccess = false;
                 if ((fileChanged.getName().equals("server.xml")) && serverXmlFileParent == null && generateFeatures) {
                     // server.xml modified
+                    // TODO: revisit scenarios in which we should skip install features
                     generateFeaturesSuccess = incrementGenerateFeatures();
                 }
                 // suppress install feature warning - property must be set before calling copyConfigFolder
@@ -4330,6 +4332,8 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
      * @throws IOException creating and copying to tempConfig directory
      */
     public void copyConfigFolder(File fileChanged, File srcDir, String targetFileName, boolean generateFeaturesSuccess) throws IOException {
+        // TODO: If we need to skip installing features after a feature generation failure in this code flow,
+        // we should also consider skipping the creation a temporary config folder
         this.tempConfigPath = Files.createTempDirectory("tempConfig");
         File tempConfig = tempConfigPath.toFile();
         debug("Temporary configuration folder created: " + tempConfig);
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1789,6 +1789,7 @@ public void restartServer(boolean buildContainer) throws PluginExecutionExceptio
         if (generateFeatures) {
             //TODO: decide if we should do an optimized feature generation on a server restart
             // If we do not generate here, need to revisit compileDependenciesChanged section in DevMojo
+            // If we generate here, we need to skip install features if there is a failure
             optimizeGenerateFeatures();
```
kathrynkodama, line 1793: Do we need to skip install features if there is a failure generating features? There may still be user specified features that we would want to install.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -3995,14 +4000,17 @@ private void processFileChanges(
...
+                boolean generateFeaturesSuccess = false;
                 if (generateFeatures) {
                     // custom server.xml modified
-                    incrementGenerateFeatures();
+                    generateFeaturesSuccess = incrementGenerateFeatures();
                 }
+                // suppress install feature warning - property must be set before calling copyConfigFolder
+                System.setProperty(SKIP_BETA_INSTALL_WARNING, Boolean.TRUE.toString());
+                // copyConfigFolder will install features if new ones are detected
+                copyConfigFolder(fileChanged, serverXmlFileParent, "server.xml", generateFeaturesSuccess);
```
kathrynkodama, line 4011: Even if generate features fails, dev mode should still install user specified features and copy their configuration file to target.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4335,7 +4347,10 @@ public boolean accept(File pathname) {
             }
         }, true);
         copyFile(fileChanged, srcDir, tempConfig, targetFileName);
-        checkConfigFile(fileChanged, tempConfig);
+        // if feature generation is on, then it should succeed before installing features
+        if (!generateFeatures || generateFeaturesSuccess) {
+            checkConfigFile(fileChanged, tempConfig);
+        }
```
kathrynkodama, line 4353: May not be relevant given comments above, but it seems redundant to create the temporary configuration folder if we may be skipping `checkConfigFile`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4029,20 +4038,22 @@ private void processFileChanges(
...
+                boolean generateFeaturesSuccess = false;
+                if ((fileChanged.getName().equals("server.xml")) && serverXmlFileParent == null && generateFeatures) {
+                    // server.xml modified
+                    generateFeaturesSuccess = incrementGenerateFeatures();
+                }
                 // suppress install feature warning - property must be set before calling copyConfigFolder
                 System.setProperty(SKIP_BETA_INSTALL_WARNING, Boolean.TRUE.toString());
-                copyConfigFolder(fileChanged, configDirectory, null);
+                // copyConfigFolder will install features if new ones are detected
+                copyConfigFolder(fileChanged, configDirectory, null, generateFeaturesSuccess);
```
kathrynkodama, line 4050: same as above, even if generate features fail we would still want to attempt to install any user specified features

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: The expectation is that if we fail to generate features, it is due to conflicts somewhere and install feature will break anyway. We could investigate this further for the restart scenario (we are currently not skipping install).

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: So far, in my testing, if generate features fails, then install features fails. It seems better to leave generate features as the error message for users to deal with in this situation. The current design in the UFO is that we do not install features after a failure to generate features. We can revisit this design if necessary.

kathrynkodama: Sounds good, thank you

kathrynkodama (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: Added TODO to revisit this in the future based on the decision for when to skip install features.

PR #335, opened 2022-01-11
Refactor the method which calls the binary scanner into two pieces for clarity and ease of maintenance

turkeylurkey:
Fixes https://github.com/OpenLiberty/ci.maven/issues/1344

I think the recursive nature of the first effort was too complex. Splitting the method in two is a little repetitive but well worth it for simplicity and maintainability.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
@@ -17,6 +17,7 @@ package io.openliberty.tools.common.plugins.util;
 
 import java.io.File;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -162,6 +132,48 @@ public BinaryScannerUtil(File binaryScanner) {
         return featureList;
     }
 
+    public Set<String> reRunBinaryScanner(Set<String> allClassesDirectories, String eeVersion, String mpVersion)
```
kathrynkodama, line 135: Perhaps a comment here to indicate how this run differs from `runBinaryScanner()`

kathrynkodama (approved):

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: ok

PR #340, opened 2022-01-18
Add support for RequiredFeatureModifiedException thrown by the binary scanner

turkeylurkey:


src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -107,9 +110,12 @@ public BinaryScannerUtil(File binaryScanner) {
                 // A RuntimeException means the currentFeatureSet contains conflicts.
                 // A FeatureConflictException means the binary files scanned conflict with each other or with
                 // the currentFeatureSet parameter.
+                // A RequiredFeatureModifiedException means the scanner can make a working list of features but
+                // only if certain inputs are changed.
                 Throwable scannerException = ite.getCause();
                 if (scannerException instanceof RuntimeException) {
-                    // The list of features from the app is passed in but it contains conflicts 
+                    // The list of features from the app is passed in but it contains conflicts
+                    // TODO have the scanner team change this to a defined exception like the others
```
kathrynkodama, line 118: Should we open an issue to track this (if we do not have one already)

kathrynkodama (approved):

kathrynkodama (approved):

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: Opened issue https://github.ibm.com/was-migration-tools/wamt/issues/4975 for this.

PR #342, opened 2022-01-25
Binary scanner, return empty array when class directories do not exist rather than an error

kathrynkodama:
Part of https://github.com/OpenLiberty/ci.maven/issues/1321 and https://github.com/OpenLiberty/ci.gradle/issues/665

Applies to the "optimize=true" scenario.
Rather than throwing an error when class directories do not exist, return an empty string. Then the binary scanner will be called with an empty string for `binaryInputs`, so that it will still detect conflicts between user specified features.


turkeylurkey (approved):
I'm glad you put the binary scanner name on line 109, I needed that once or twice.

TrevCraw (approved):

PR #348, opened 2022-03-10
Call the binary scanner using the new API: new names, log location and log level

turkeylurkey:
Once this is merged we'll need to install the jar to the local maven repository until the jar is uploaded to sonatype.

Signed-off-by: Paul Gooderham <turkeyonmarblerye@gmail.com>

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
@@ -70,6 +70,7 @@ public abstract class BinaryScannerUtil {
     public abstract void error(String message);
     public abstract void warn(String message);
     public abstract void info(String message);
+    public abstract boolean isDebugEnabled();
 
     // The jar file containing the binary scanner code
     private File binaryScanner;
@@ -112,7 +113,10 @@ public abstract class BinaryScannerUtil {
                 Method generateFeatureSetMethod = getScannerMethod();
                 // names: binaryInputs, targetJavaEE, targetMicroProfile, currentFeatures, logLocation, logLevel, locale
                 Set<String> binaryInputs = getBinaryInputs(classFiles, allClassesDirectories, optimize);
-                String logLevel = "*=FINE"; // generate messages for debugging by support team
+                String logLevel = null;
+                if (isDebugEnabled()) {
+                    logLevel = "*=FINE";  // generate messages for debugging by support team
+                }
                 debug("Calling " + binaryScanner.getName() + " with the following inputs...\n" +
                         "  binaryInputs: " + binaryInputs + "\n" +
                         "  eeVersion: " + eeVersion + "\n" +
--- a/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
@@ -209,9 +209,12 @@ public abstract class BinaryScannerUtil {
             Method generateFeatureSetMethod = getScannerMethod();
             Set<String> binaryInputs = allClassesDirectories;
             Set<String> currentFeaturesSet = new HashSet<String>(); // when re-running always pass in no features
-            String logLevel = null;
+            String logLevel;
             if (isDebugEnabled()) {
                 logLevel = "*=FINE";  // generate messages for debugging by support team
+            } else {
+                logLevel = null;
+                logLocation = null;
             }
             debug("Recalling binary scanner with the following inputs...\n" +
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -102,62 +104,59 @@ public BinaryScannerUtil(File binaryScanner) {
...
         Set<String> featureList = null;
         if (binaryScanner != null && binaryScanner.exists()) {
             try {
-                Method driveScanMavenFeatureList = getScannerMethod();
-                String[] binaryInputs = getBinaryInputs(classFiles, allClassesDirectories, optimize);
-                List<String> currentFeatures = new ArrayList<String>(currentFeatureSet);
+                Method generateFeatureSetMethod = getScannerMethod();
+                // names: binaryInputs, targetJavaEE, targetMicroProfile, currentFeatures, logLocation, logLevel, locale
+                Set<String> binaryInputs = getBinaryInputs(classFiles, allClassesDirectories, optimize);
+                String logLevel = "*=FINE"; // generate messages for debugging by support team
```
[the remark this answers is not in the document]
turkeylurkey: If you use "*=INFO" there are few messages. Do we want traceability by default or just when -X debug is used?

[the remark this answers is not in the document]
turkeylurkey: Ok, sounds good.

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -102,62 +105,65 @@ public BinaryScannerUtil(File binaryScanner) {
...
+                }
                 debug("Calling " + binaryScanner.getName() + " with the following inputs...\n" +
-                      "  binaryInputs: " + Arrays.toString(binaryInputs) + "\n" +
-                      "  eeVersion: " + eeVersion + "\n" +
-                      "  mpVersion: " + mpVersion + "\n" +
-                      "  currentFeatures: " + currentFeatures + "\n" +
-                      "  locale: " + java.util.Locale.getDefault());
+                        "  binaryInputs: " + binaryInputs + "\n" +
+                        "  eeVersion: " + eeVersion + "\n" +
+                        "  mpVersion: " + mpVersion + "\n" +
```
kathrynkodama, line 126: For consistency with the comment above could rename to "targetJavaEE" and "targetMicroProfile"

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -102,62 +105,65 @@ public BinaryScannerUtil(File binaryScanner) {
...
+                        "  currentFeatures: " + currentFeatureSet + "\n" +
+                        "  logLocation: " + logLocation + "\n" +
+                        "  logLevel: " + logLevel + "\n" +
+                        "  locale: " + java.util.Locale.getDefault());
                 debug("The following messages are from the application binary scanner used to generate Liberty features");
```
kathrynkodama, line 131: What sort of messages can be returned from the binary scanner that we do not catch in the exceptions below or in the logs?

kathrynkodama (approved):
LGTM

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: In prior versions the binary scanner could print messages. In the second most recent jar it would print messages if there were no logs directory but the latest jar has fixed this. Do you think we should remove these debug statements now?

kathrynkodama: Thanks Paul, in that case I think these statements should be removed.

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: Good idea, will do.

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: Ok

TrevCraw:
Fixes https://github.com/OpenLiberty/ci.maven/issues/1443 and https://github.com/OpenLiberty/ci.maven/issues/1385
Part of https://github.com/OpenLiberty/ci.maven/issues/1380

PR #352, opened 2022-04-07
Clean up console output for dev mode and generate features from ServerFeatureUtil

kathrynkodama:
See https://github.com/OpenLiberty/ci.maven/pull/1477 and https://github.com/OpenLiberty/ci.gradle/pull/729

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -4374,8 +4374,8 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
      * @throws IOException creating and copying to tempConfig directory
      */
     public void installFeaturesToTempDir(File fileChanged, File srcDir, String targetFileName, boolean generateFeaturesSuccess) throws IOException {
-        if (generateFeatures && !generateFeaturesSuccess) {
-            return; // skip creating temp dir and installing features if feature generation failed
+        if ((generateFeatures && !generateFeaturesSuccess) || !serverFeaturesModified()) {
+            return; // skip creating temp dir and installing features if feature generation failed or server features have not changed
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -5419,9 +5426,9 @@ private boolean classesFailingToCompile() {
     }
 
     // Returns true if features have been modified in the configuration directory
+    // (excluding generated-features.xml file). Called only if generate-features=true
```
TrevCraw, line 5429: Curious about this comment. It seems like `serverFeaturesModified()` is called on line 4030 regardless of the `generateFeatures` flag value. It also seems like the return value from this method is intended to determine whether or not we should install features regardless of `generateFeatures` value. Is this the intention?

TrevCraw:
I think both these lines should be changed to `debug`:

https://github.com/OpenLiberty/ci.common/blob/bef758b339b5bfb501e851aff653ec453b353878/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java#L2467
https://github.com/OpenLiberty/ci.common/blob/bef758b339b5bfb501e851aff653ec453b353878/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java#L2483

turkeylurkey (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
kathrynkodama: You're correct it is no longer used only in the geenrateFeatures scenario, now we use it outside of that scenario to avoid creating and copying to a temp dir if the server features have not changed. I've updated the comment

TrevCraw (approved):

PR #361, opened 2022-04-27
1469 update class file tracking

TrevCraw:
Fixes https://github.com/OpenLiberty/ci.maven/issues/1469

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2843,7 +2843,6 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                 // Assumption is that initialCompile will only be set to false after ALL modules are complete
                 // Will need to update logic once multiple attempts are enabled for upstream projects
                 if (!recompileDependencies && !initialCompile && !sourceOutputDirRegistered) {
-                    debug("Attempt to register class file directories for feature generation");
                     // Allow more than one register attempt for adding a new output directory once adding a src/main/java dir
                     // to an upstream project is supported - https://github.com/OpenLiberty/ci.maven/issues/1202
                     if (isMultiModuleProject() && !mmOutputDirRegAttempted) {
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -4001,9 +4001,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                             debug("Multi-module: Setting modifiedSrcBuildFile to: " + modifiedSrcBuildFile);
                             ProjectModule modifiedModule = getProjectModule(modifiedSrcBuildFile);
                             File outputDir = modifiedModule.getOutputDirectory();
-                            if (!javaSourceClasses.contains(outputDir)) {
-                                javaSourceClasses.add(outputDir);
-                            }
+                            javaSourceClasses.add(outputDir);
                             lastChangeCompiled = false;
                             triggerUpstreamModuleCompile(project, false);
                         } else {
@@ -4115,9 +4113,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                     // set to know what module to generate features for
                     modifiedSrcBuildFile = buildFile;
                     debug("Single module: Setting modifiedSrcBuildFile to: " + modifiedSrcBuildFile);
-                    if (!javaSourceClasses.contains(outputDirectory)) {
-                        javaSourceClasses.add(outputDirectory);
-                    }
+                    javaSourceClasses.add(outputDirectory);
                     lastChangeCompiled = false;
                     triggerMainModuleCompile(false);
                 } else {
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2618,8 +2618,6 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
     File dockerfileUsed;
     File looseAppFile;
     WatchService watcher;
-    boolean mainModuleCompiled;
-    boolean upstreamModuleCompiled;
     boolean lastChangeCompiled;
 
     // used for multi module projects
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2849,26 +2842,55 @@ public void watchFiles(File outputDirectory, File testOutputDirectory, final Thr
...
-                            if (isMultiModuleProject()) {
-                                runTestThread(false, executor, -1, false, getAllBuildFiles());
-                            } else {
-                                runTestThread(false, executor, -1, false, false, buildFile);
+                // Assumption is that initialCompile will only be set to false after ALL modules are complete
+                if (!recompileDependencies && !registerClassDirAttempted && !initialCompile) {
+                    debug("Attempt to register class file directories for feature generation");
+                    // set to true regardless of success
+                    // this may need to be updated, but currently do not want to come into this section again
+                    registerClassDirAttempted = true;
```
kathrynkodama, line 2850: Why do we only want to enter this block once? Will we catch the case where java classes are added after dev mode runs for a while, and then we register the output directory?

Rather than the `registerClassDirAttempted` boolean, instead can we use the `sourceOutputDirRegistered` existing booleans similar to how we do for `sourceDirRegistered` once we are in the watch loop?
ie. 
```java
if (!recompileDependencies && !initialCompile){
    if (isMultiModuleProject()) {
        for (ProjectModule p : upstreamProjects) {
              if (!p.sourceOutputDirRegistered && p.getOutputDirectory.exists() && shouldIncludeSources(p.getPackagingType())) {
                  // register  
               }
        }
    }
    if (shouldIncludeSources(packagingType) && !sourceOutputDirRegistered && this.outputDirectory.exists()) {
        // register 
    }
}
```

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2849,26 +2842,55 @@ public void watchFiles(File outputDirectory, File testOutputDirectory, final Thr
...
+                    }
+                    // TODO: look into whether or not we need this code block
+                    if (tryFeatureGeneration && !generatedFeaturesFile.exists()) {
+                        // run tests if generated-features.xml does not exist as there are no new features to install
+                        if (isMultiModuleProject()) {
+                            runTestThread(false, executor, -1, false, getAllBuildFiles());
+                        } else {
+                            runTestThread(false, executor, -1, false, false, buildFile);
+                        }
+                    }
```
kathrynkodama, line 2893: I think we still need this block of code. If dev mode starts and the server.xml already contains all of the features needed, we will not create a generated-features.xml file. This catches the case where `hotTests=true`, we want to run tests after feature generation finishes.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2698,6 +2703,7 @@ public void watchFiles(File outputDirectory, File testOutputDirectory, final Thr
             // check for upstream projects
             if (isMultiModuleProject()) {
                 for (ProjectModule p : upstreamProjects) {
+                    // TODO: Consider whether or not we should generate features at this point
```
kathrynkodama, line 2706: Why would we generate features here?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4240,7 +4284,7 @@ && serverFeaturesModified()) {
         } else if (isDockerfileDirectoryChanged(fileChanged)) {
             // If contents within a directory specified in a Dockerfile COPY command were changed, and not already processed by one of the other conditions above.
             untrackDockerfileDirectoriesAndRestart();
-        } else if (generateFeatures && directory.startsWith(srcOutputPath)) {
+        } else if (!recompileDependencies && generateFeatures && directory.startsWith(srcOutputPath)) {
```
kathrynkodama, line 4287: Suggestion to add some comments indicating why we track class files when `recompileDependencies=false` and what the new behaviour is when `recompileDependencies=true`

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2616,6 +2618,7 @@ private void readInput() {
     File dockerfileUsed;
     File looseAppFile;
     WatchService watcher;
+    boolean lastChangeCompiled;
```
kathrynkodama, line 2621: Suggestion to add some comments indicating what lastChangeCompiled is tracking/how it is used.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2849,26 +2842,55 @@ public void watchFiles(File outputDirectory, File testOutputDirectory, final Thr
...
+                        debug("Detected a change in the following class directories: " + javaSourceClasses);
+                        lastChangeCompiled = false;
+                        modifiedSrcBuildFile = null;
+                        tryFeatureGeneration = true;
+                        incrementGenerateFeatures();
+                    } else if (!recompileDependencies && !javaSourceClasses.isEmpty()) { // class file tracking
+                        debug("Detected a change in the following classes: " + javaSourceClasses);
+                        tryFeatureGeneration = true;
+                        incrementGenerateFeatures();
+                    }
```
kathrynkodama, line 2884: I think this could be simplified to remove the `tryFeatureGeneration` boolean (just a suggestion): 
```java
if (generateFeatures && !classesFailingToCompile && !javaSourceClasses.isEmpty()  {
    if ((recompileDependencies && lastChangeCompiled) || !recompileDependencies) {
        if (recompileDependencies) {
            // no class file tracking
            debug("Detected a change in the following class directories: " + javaSourceClasses);
            lastChangeCompiled = false;
            modifiedSrcBuildFile = null;
        } else {
             // class file tracking
            debug("Detected a change in the following classes: " + javaSourceClasses);
        }
        incrementGenerateFeatures();
        if (!generatedFeaturesFile.exists()) {
             // run tests logic
        }
    }
}
```

turkeylurkey (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: I am thinking that we probably do not want generate features at this point, but the way the code is currently written, we may re-generate features inside updateArtifacts if compilation dependencies have changed. I am not sure if that is the behaviour we want at start up or not, which is why I added the TODO.

kathrynkodama: Thanks, that makes sense! I agree we do not want to generate features. I think it is safe to leave that logic in updateArtifacts for now in the off chance that the dependencies have changed since dev mode startup.

kathrynkodama (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
TrevCraw: Updated to only have one block where we attempt to register the output directories.

PR #362, opened 2022-04-29
Do not rerun binary scanner if optimizing and there are no user specified features

kathrynkodama:
Part of https://github.com/OpenLiberty/ci.maven/issues/1497

When a `ProvidedFeatureConflictException`, `FeatureConflictException` or `RequiredFeatureModifiedException` is returned from the binary scanner, we do not need to call it again if we have already generated features for all class files (optimize = true) and we did not pass any user specified features (currentFeatureSet is empty).

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
@@ -172,7 +172,8 @@ public abstract class BinaryScannerUtil {
                 if (scannerException.getClass().getName().equals(PROVIDED_FEATURE_EXCEPTION)) {
                     // The list of features from the app is passed in but it contains conflicts
                     Set<String> conflicts = getFeatures(scannerException);
-                    Set<String> sampleFeatureList = reRunIfFailed ? reRunBinaryScanner(allClassesDirectories, logLocation, targetJavaEE, targetMicroProfile): null;
+                    // always rerun binary scanner in this scenario, this exception only occurs if a current feature list is passed to binary scanner
+                    Set<String> sampleFeatureList = reRunBinaryScanner(allClassesDirectories, logLocation, targetJavaEE, targetMicroProfile);
                     if (sampleFeatureList == null) {
                         throw new NoRecommendationException(conflicts);
                     } else {
@@ -181,6 +182,7 @@ public abstract class BinaryScannerUtil {
                 } else if (scannerException.getClass().getName().equals(FEATURE_CONFLICT_EXCEPTION)) {
                     // The scanned files conflict with each other or with current features
                     Set<String> conflicts = getFeatures(scannerException);
+                    //  rerun binary scanner with all class files and without the current feature set to get feature recommendations
                     Set<String> sampleFeatureList = reRunIfFailed ? reRunBinaryScanner(allClassesDirectories, logLocation, targetJavaEE, targetMicroProfile): null;
                     if (sampleFeatureList == null) {
                         throw new NoRecommendationException(conflicts);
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -168,7 +172,7 @@ public Set<String> runBinaryScanner(Set<String> currentFeatureSet, List<String>
                 if (scannerException.getClass().getName().equals(PROVIDED_FEATURE_EXCEPTION)) {
                     // The list of features from the app is passed in but it contains conflicts
                     Set<String> conflicts = getFeatures(scannerException);
-                    Set<String> sampleFeatureList = reRunBinaryScanner(allClassesDirectories, logLocation, targetJavaEE, targetMicroProfile);
+                    Set<String> sampleFeatureList = reRunIfFailed ? reRunBinaryScanner(allClassesDirectories, logLocation, targetJavaEE, targetMicroProfile): null;
```
TrevCraw, line 175: I think we should use the `featureList` returned from the first call to the binary scanner instead of `null`.
Same for the other two reruns below.

turkeylurkey: I'm not so sure. If the message is "A working set of features could not be generated due to conflicts in the application’s API usage: A, B, C" you can't say "...suggested features for guidance: A, B, C."

TrevCraw: For the `FEATURE_CONFLICT_EXCEPTION`, I agree. I will add a comment there.

For `PROVIDED_FEATURE_EXCEPTION`, we actually do not need to add this extra rerun check logic because this exception only occurs if a feature list was passed in, which means we cannot skip the second binary scanner call.

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -177,7 +181,7 @@ public Set<String> runBinaryScanner(Set<String> currentFeatureSet, List<String>
                 } else if (scannerException.getClass().getName().equals(FEATURE_CONFLICT_EXCEPTION)) {
                     // The scanned files conflict with each other or with current features
                     Set<String> conflicts = getFeatures(scannerException);
-                    Set<String> sampleFeatureList = reRunBinaryScanner(allClassesDirectories, logLocation, targetJavaEE, targetMicroProfile);
+                    Set<String> sampleFeatureList = reRunIfFailed ? reRunBinaryScanner(allClassesDirectories, logLocation, targetJavaEE, targetMicroProfile): null;
```
TrevCraw, line 184: As Paul mentioned in https://github.com/OpenLiberty/ci.common/pull/362/files#r862254383, the error here (in the case of an "optimize" failure), would always be "A working set of features could not be generated due to conflicts in the application’s API usage", hence the sampleFeatureList will always be `null`. So this line is good as is.

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -186,7 +190,7 @@ public Set<String> runBinaryScanner(Set<String> currentFeatureSet, List<String>
                 } else if (scannerException.getClass().getName().equals(FEATURE_MODIFIED_EXCEPTION)) {
                     // The scanned files conflict and the scanner suggests modifying some features
                     Set<String> modifications = getFeatures(scannerException);
-                    Set<String> sampleFeatureList = reRunBinaryScanner(allClassesDirectories, logLocation, targetJavaEE, targetMicroProfile);
+                    Set<String> sampleFeatureList = reRunIfFailed ? reRunBinaryScanner(allClassesDirectories, logLocation, targetJavaEE, targetMicroProfile) : null;
```
TrevCraw, line 193: For the `FEATURE_MODIFIED_EXCEPTION` case, we should use the `featureList` from the first call as the `sampleFeatureList` if we are not making a second call, as we will want this list either recommended or applied depending on the scenario.

TrevCraw: oh I see, I thought the `sampleFeatureList` was used for the recommendation/apply logic. Since it is not, I agree, we should use null.

TrevCraw (approved):

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
kathrynkodama: I've updated the logic here to remove the rerun check logic.

turkeylurkey (approved):

PR #381, opened 2022-06-27
Compose version parameters for binary scanner and handle errors

turkeylurkey:
Instead of hard coding the parameters passed to the binary scanner we will compose the parameter from the information in the build file. If the build file specifies a version number that the binary scanner does not support it will throw an invalid argument exception which we report.

This pull request is also a bit of a design review. We could request a better API from the binary scanner regarding invalid version numbers but the current arrangement looks pretty safe. Also the message we report should be reviewed.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
@@ -388,7 +388,7 @@ public abstract class BinaryScannerUtil {
      * E.g. mp1.3, mp4.1 etc
      * @param ver the String value version number read from the build file (pom.xml, build.gradle)
      *           E.g. 1, 2.1 etc. This is verified by the parser and cannot be blank.
-     * @return String parameter passed to binary scanner or empty string in case of error
+     * @return String parameter passed to binary scanner or null in case of error
      */
     public static String composeMPVersion(String ver) {
         int offset = ver.indexOf("-RC"); // clean up 4.1-RC to 4.1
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -206,10 +182,21 @@ public Set<String> runBinaryScanner(Set<String> currentFeatureSet, List<String>
...
-                    throw new PluginExecutionException("Error scanning the application for Liberty features: " + scannerException.toString());
+                } else if (scannerException.getClass().getName().contains("java.lang.IllegalArgumentException")) {
+                    String msg = scannerException.getMessage();
+                    if (msg.contains("CWMIG12056E")) {
+                        if (msg.contains("targetJavaEE")) {
+                            throw new PluginExecutionException("The Java EE or Jakarta EE version number specified in the build file " +
+                                    "is not supported for feature generation.");
+                        } else if (msg.contains("targetMicroProfile")) {
+                            throw new PluginExecutionException("The MicroProfile version number specified in the build file " +
+                                    "is not supported for feature generation.");
```
kathrynkodama, line 193: Suggestion to define these messages as constants so that they can be referenced in the integration test without defining new constants.
See https://github.com/OpenLiberty/ci.maven/blob/8b3da42bd86195596d2d7dddabba840b951d9691/liberty-maven-plugin/src/it/generate-features-it/src/test/java/net/wasdev/wlp/test/dev/it/GenerateFeaturesTest.java#L180-L182

kathrynkodama (approved):

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: Good idea.

PR #384, opened 2022-08-19
Disable docker scan suggestion on devc docker build

scottkurz:
Signed-off-by: Scott Kurz <skurz@us.ibm.com>

Fixes https://github.com/OpenLiberty/ci.maven/issues/1575

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1354,12 +1354,16 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
     private Process getRunProcess(String command) throws IOException {
         ProcessBuilder processBuilder = new ProcessBuilder();
         processBuilder.command(getCommandTokens(command));
+        Map<String, String> env = processBuilder.environment();
         if (!OSUtil.isLinux()){
-            Map<String, String> env = processBuilder.environment();
             if (!env.keySet().contains("DOCKER_BUILDKIT")) { // don't touch if already set
                 env.put("DOCKER_BUILDKIT", "0"); // must set 0 on Windows VMs
                 debug("Generating environment for docker build & run: DOCKER_BUILDKIT=0");
             }
+        }
+        if (!env.keySet().contains("DOCKER_SCAN_SUGGEST")) { // don't touch if already set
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1354,13 +1354,17 @@ private void startContainer() {
...
+        Map<String, String> env = processBuilder.environment();
         if (!OSUtil.isLinux()){
-            Map<String, String> env = processBuilder.environment();
             if (!env.keySet().contains("DOCKER_BUILDKIT")) { // don't touch if already set
                 env.put("DOCKER_BUILDKIT", "0"); // must set 0 on Windows VMs
                 debug("Generating environment for docker build & run: DOCKER_BUILDKIT=0");
             }
         }
+        if (!env.keySet().contains("DOCKER_SCAN_SUGGEST")) { // don't touch if already set
+            env.put("DOCKER_SCAN_SUGGEST", "false"); // Disable stderr contents from showing as error message
```
TrevCraw, line 1365: Just to be clear, the intention is not to disable all stderr contents from being displayed as an error message, right?
Just the "Synk tests" message?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1354,13 +1354,17 @@ private void startContainer() {
...
+            debug("Generating environment for docker build & run: DOCKER_SCAN_SUGGEST=false");
```
cherylking, line 1366: If we are on a non-Linux system, are we going to get two "Generating environment for docker build & run:" messages? The one at line 1361 and the one here?

cherylking: I guess it's just debug messages. So that should be fine.

cherylking (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
scottkurz: Right, the general idea of propagating stderr to error logging msgs remains intact...we just stopped this particular msg getting written to stderr by `docker build`.    Yes, it seems to be also doing the same for `docker run` but the DOCKER_BUILDKIT one did too so I guess that's OK.

TrevCraw (approved):

PR #385, opened 2022-08-22
Add support for the binary scanner checking the validity of the target versions for ee and mp

turkeylurkey:
This support will be added in the 22.0.0.4 (or later) version of the binary scanner and this code continues to work with the 22.0.0.3 version which is currently available.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
@@ -412,6 +412,9 @@ public abstract class BinaryScannerUtil {
      * @return String parameter passed to binary scanner
      */
     public static String composeEEVersion(String ver) {
+        if (ver == null) {
+            return null;
+        }
         String majorVersion;
         int offset = ver.indexOf(".");
         majorVersion = (offset == -1) ? ver : ver.substring(0, offset);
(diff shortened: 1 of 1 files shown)


TrevCraw (approved):

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -488,4 +522,38 @@ public String getMPLevel() {
...
+        public String getEELevel() {
+            return eeLevel;
+        }
+        public String getMPLevel() {
+            return mpLevel;
+        }
+    }
+
+    // A class to pass the EE and MP parameters which do not work together back to the caller.
+    public class IllegalTargetComboException extends Exception {
```
kathrynkodama, line 544: (Suggestion) This could extend the `IllegalTargetException` class above  since the code looks identical to get the `eeLevel` and `mpLevel`

kathrynkodama (approved):

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: Good idea. I like to have separate exceptions to represent the two different concepts but no point duplicating all the code.

kathrynkodama (approved):

PR #386, opened 2022-08-23
write liberty-dev-metadata when using devc

dshimo:
Screenshot demonstrated with the liberty starter-app

File is written out to `{buildDir}/{serverName}-liberty-devc-metadata.xml`
The original file gets replaced if/when written to the same place when stopped.

image

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -401,8 +401,6 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
     private final boolean container;
     private String imageName;
     private String containerName;
-    protected String containerStartTimestamp;
-    protected String containerStopTimestamp;
     private File dockerfile;
     private File dockerBuildContext;
     private Path tempDockerfilePath = null;
@@ -1395,8 +1393,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
         });
         logCopyErrorThread.start();
 
-        containerStartTimestamp = Timestamp.from(Instant.now()).toString();
-        writeDevcMetadata();
+        writeDevcMetadata(true);
         if (timeout == 0) {
             startingProcess.waitFor();
         } else {
@@ -1411,8 +1408,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
             debug("Unexpected exit running docker command, return value=" + startingProcess.exitValue());
             // show first message from standard err
             String errorMessage = new String(firstErrorLine).trim() + " RC=" + startingProcess.exitValue();
-            containerStopTimestamp = Timestamp.from(Instant.now()).toString();
-            writeDevcMetadata();
+            writeDevcMetadata(false);
             throw new RuntimeException(errorMessage);
         }
     }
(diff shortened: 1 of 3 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -5638,4 +5653,37 @@ private Set<String> getGeneratedFeatures() {
...
+    /**
+     * Create metadata when running devc mode and containers
+     * Language server then uses metadata file to connect
+     * 
+     * @throws XMLStreamException
+     * @throws IOException
+     * @throws FactoryConfigurationError
+     */
+    public void writeDevcMetadata() {
+        File metaFile = new File(serverDirectory, "liberty-dev-metadata.xml");
```
cherylking, line 5666: As discussed in Slack, let's use the build directory instead of the server directory for the location of this file, and simply add the name of the server to the file name.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -5638,4 +5653,37 @@ private Set<String> getGeneratedFeatures() {
...
+        try {
+            XMLStreamWriter metadataWriter = XMLOutputFactory.newInstance().createXMLStreamWriter(new FileWriter(metaFile)) ;
+            metadataWriter.writeStartDocument();
+            metadataWriter.writeStartElement("devModeMetaData");
```
cherylking, line 5670: This file only exists for `devc` mode right? So maybe change this element to `devcModeMetaData`.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1386,6 +1395,8 @@ public void run() {
         });
         logCopyErrorThread.start();
 
+        containerStartTimestamp = Timestamp.from(Instant.now());
+        writeDevcMetadata();
```
cherylking, line 1399: Would you ever call `writeDevcMetadata` without first setting a start or stop time immediately before? It seems like they always go together. So maybe combine them in one method? You could have a `createDevcMetadata(boolean start)` and if the `start` flag is true create it with a start timestamp, otherwise create it with a stop timestamp.  Actually, the timestamp seems irrelevant. We never actually look at the value to determine something. It's just whether it is there or not.

src/test/java/io/openliberty/tools/common/plugins/util/DevUtilTest.java
```diff
@@ -527,5 +530,14 @@ public void testOmitFilesList() throws Exception {
...
     }
 
-
+    @Test
+    public void testWriteDevcMetadataSimple() throws XMLStreamException, FactoryConfigurationError, IOException {
+        util.writeDevcMetadata();
+        File metaDataXml = new File(serverDirectory, "liberty-dev-metadata.xml");
+        assertTrue(metaDataXml.exists());
+        String content = FileUtils.readFileToString(metaDataXml, "UTF-8");
+        // assertTrue(content.contains("<startTime>"));
```
cherylking, line 539: Why is this commented out? If the `<stopTime>` below is not present, shouldn't the `<startTime>` be present? One or the other should always be there?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -5638,4 +5653,34 @@ private Set<String> getGeneratedFeatures() {
...
+     * 
+     */
+    public void writeDevcMetadata() {
+        File metaFile = new File(buildDirectory, serverDirectory.getName() + "-liberty-devc-metadata.xml");
+        try {
+            XMLStreamWriter metadataWriter = XMLOutputFactory.newInstance().createXMLStreamWriter(new FileWriter(metaFile)) ;
+            metadataWriter.writeStartDocument();
+            metadataWriter.writeStartElement("devModeMetaData");
+            writeElement(metadataWriter, "containerName", containerName != null ? containerName : DEVMODE_CONTAINER_BASE_NAME);
+            if (containerStartTimestamp != null) writeElement(metadataWriter, "startTime", containerStartTimestamp.toString());
```
cherylking, line 5669: You changed these vars to Strings already. So no need to call `toString()` on them.

cherylking (approved):

PR #392, opened 2023-02-06
fix debug trace for prepareFeature goal

jjiwooLim:
The key for getting exception should be to `action.exception.stacktrace`

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/PrepareFeatureUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/PrepareFeatureUtil.java
@@ -251,8 +251,10 @@ public abstract class PrepareFeatureUtil extends ServerFeatureUtil {
 				json = (File) mapBasedInstallKernel.get("generate.json");
 				
 				if (mapBasedInstallKernel.get("action.error.message") != null) {
+					debug("generateJson action.error.message: "
+							+ mapBasedInstallKernel.get("action.error.message"));
 					debug("generateJson action.exception.stacktrace: "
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/PrepareFeatureUtil.java
```diff
@@ -251,8 +251,10 @@ public File generateJson(String targetJsonFile, Map<File, String> esaFileMap) th
 				json = (File) mapBasedInstallKernel.get("generate.json");
 				
 				if (mapBasedInstallKernel.get("action.error.message") != null) {
+					debug("generateJson action.error.message: "
+							+ mapBasedInstallKernel.get("action.error.message"));
 					debug("generateJson action.exception.stacktrace: "
-							+ mapBasedInstallKernel.get("action.error.stacktrace"));
+							+ mapBasedInstallKernel.get("action.exception.stacktrace"));
```
cherylking, line 257: We have to work with various different versions of Liberty. If this key changed in a recent version, seems like this code would need to handle both keys?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/PrepareFeatureUtil.java
jjiwooLim: Sorry I should have mentioned this is actually a bug. The key has been 'action.exception.stacktrace' for a long period of time. I will check for any additional references to this key.

cherylking (approved):

PR #402, opened 2023-06-05
Check for empty feature

cherylking:
One more place to check for empty feature, issue warning and skip adding to the collection.


scottkurz (approved):
LGTM

PR #403, opened 2023-06-08
Make handling of temp config dir better and tolerate exceptions during shutdown/cleanup

cherylking:


cherylking:
@TrevCraw Since the file that is locked is in the `messaging\messageStore\Log` directory, should we just skip `messaging` which would skip everything under that directory?

Screenshot 2023-06-08 at 3 17 29 PM

TrevCraw (approved):

Issue #404, opened 2023-06-13
maven-artifact should be "provided"


bmarwell:
Maven artifact should be "provided".
The version is not synchronized among other projects, e.g. ci.maven. I would recommend to extact its version into a property which can then be consumed by ci.maven.

cherylking:
After further discussion, I think this dependency should be removed from ci.common. Given that our Gradle plugin also depends on this repo, having a dependency on a Maven artifact does not make sense. 

I will open a PR to remove it. The only classes we were utilizing were `DefaultArtifactVersion` and `ComparableVersion` to compare versions. We can do that in a different way.

PR #405, opened 2023-06-13
[#404] fix scope for maven-artifact

bmarwell:
* extract version into property to be consumed by ci.maven
* Scope is prerequisite for https://github.com/OpenLiberty/ci.maven/issues/1691
* extraction is sensible for using the same version in ci.maven.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,6 +35,8 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <maven.compiler.source>1.7</maven.compiler.source>
         <maven.compiler.target>1.7</maven.compiler.target>
+
+        <dependency.maven-artifact.version>3.6.3</dependency.maven-artifact.version>
     </properties>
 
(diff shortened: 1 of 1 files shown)


cherylking:
@bmarwell I know I already merged this, but wanted to check real quick...will this negatively affect the Liberty Gradle plugin (ci.gradle) which also utilizes this code? If you are not sure, then I probably should pull this change in locally and make sure the gradle plugin builds and tests ok.

bmarwell:
Argh. I honestly have no idea about Gradle.
Maven ships it and emits a warning if present. Gradle doesn't ship it which breaks maven compat, it seems. I guess it's either importing explicitly into gradle (I'd prefer that) or exclude in maven explicitly.

cherylking:
I'm working on a separate PR to remove this dependency from ci.common altogether. Given it is used by both our Maven and Gradle plugins, it does not make much sense to have a dependency on something provided by Maven IMO.

PR #409, opened 2023-08-11
Avoid unnecessary war:exploded calls for mere recompilation of Java c…

scottkurz:
…lasses

Prereq for https://github.com/OpenLiberty/ci.maven/pull/1714.

Fixes:
1. Exposed filtered resource/webResource dirs for LMP/LGP (LMP will issue message explaining it can't update this list)
2. Canonicalize path for proj dir
3. Avoid calling updateLooseApp() (which does another war:exploded call) for simply a Java compilation (class update)

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -3894,7 +3894,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
         Path configPath = this.configDirectory.getCanonicalFile().toPath();
         Path outputPath = this.outputDirectory.getCanonicalFile().toPath();
 
-        Path directory = fileChanged.getParentFile().toPath();
+        Path directory = fileChanged.getParentFile().getCanonicalFile().toPath();
 
         // resource file check
         File resourceParent = null;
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -3893,7 +3894,7 @@ private void processFileChanges(
         Path configPath = this.configDirectory.getCanonicalFile().toPath();
         Path outputPath = this.outputDirectory.getCanonicalFile().toPath();
 
-        Path directory = fileChanged.getParentFile().toPath();
+        Path directory = fileChanged.getParentFile().getCanonicalFile().toPath();
```
cherylking, line 3897: Curious about these changes because we use `toPath()` on File objects in LCLS. Why did you need to add the call to `getCanonicalFile()`?

cherylking (approved):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
cherylking: Discussed offline that this makes it consistent with the other lines of code in this method.

scottkurz: This got me further in doing the loose app IT for something like this...
 >    <dir sourceOnDisk="/tmp/temp6307480008441095444/target/classes" targetInArchive="/WEB-INF/classes"/>
on Windows.. it would sometimes use a Windows-ism that made it hard to match (I don't have an example handy.. but the thing where it shortens paths to 8 chars... like Admini~1 or something.

I ended up using a different approach though in IT, so I guess the rationalization is it now looks more like the other paths, and it passed the GHA test runs.

PR #411, opened 2023-09-06
Update container functions to use Podman

mattbsox:
Adding feature to detect Podman installations and use it to run dev mode with podman.
- Will default to Docker if both are present
- Added support for Containerfiles

PRs with doc updates and tests for LMP/LGP to follow.


src/main/java/io/openliberty/tools/common/plugins/util/AbstractContainerSupportUtil.java
```diff
@@ -7,6 +23,14 @@
 
 public abstract class AbstractContainerSupportUtil {
 
+    private static final String CONTAINER_DOCKER_PREFIX = "docker ";
+    private static final String CONTAINER_PODMAN_PREFIX = "podman ";
+
+    private static final int DOCKER_TIMEOUT = 20; // seconds
```
cherylking, line 29: Is this timeout being used for both container types? If so, make name generic.

src/main/java/io/openliberty/tools/common/plugins/util/AbstractContainerSupportUtil.java
```diff
@@ -21,21 +45,81 @@ public abstract class AbstractContainerSupportUtil {
...
+     */
+    private static final String MIN_DOCKER_VERSION = "18.03.0"; // Must use Docker 18.03.0 or higher
+    protected void checkDockerVersion() throws PluginExecutionException {
+        String versionCmd = "docker version --format {{.Client.Version}}";
+        String dockerVersion = execContainerCmd(versionCmd, DOCKER_TIMEOUT);
+        if (dockerVersion == null) {
+            checkPodmanVersion(); // Check Podman version if no Docker 
+            return;
+        }
+        debug("Detected Docker version >" + dockerVersion);
```
cherylking, line 79: This looks like it is saying the Docker version is greater than what was found. I would change to "Detected Docker version: ". Same comment for the Podman version debug message.

src/main/java/io/openliberty/tools/common/plugins/util/AbstractContainerSupportUtil.java
```diff
@@ -21,21 +45,81 @@ public abstract class AbstractContainerSupportUtil {
...
+        
+        if (VersionUtility.compareArtifactVersion(dockerVersion, MIN_DOCKER_VERSION, false) < 0) {
+            checkPodmanVersion(); // Check that bad Docker version isn't just a Podman version
+            if (!isDocker) {
+                return;
+            }
+            throw new PluginExecutionException("The detected Docker client version number is not supported:" + dockerVersion.trim() + ". Docker version must be 18.03.0 or higher.");
```
cherylking, line 86: Use the constant in the exception for the minimum version in case it changes in the future. Then we only have to change in one place.

src/main/java/io/openliberty/tools/common/plugins/util/AbstractContainerSupportUtil.java
```diff
@@ -21,21 +45,81 @@ public abstract class AbstractContainerSupportUtil {
...
+    private void checkPodmanVersion() throws PluginExecutionException  {
+        String versionCmd = "podman version --format {{.Client.Version}}";
+        String podmanVersion = execContainerCmd(versionCmd, DOCKER_TIMEOUT);
+        if (podmanVersion == null) {
+            return; // Can't tell if the version is valid.
+        }
+        debug("Detected Podman version >" + podmanVersion);
+        
+        if (VersionUtility.compareArtifactVersion(podmanVersion, MIN_PODMAN_VERSION, false) < 0) {
+            throw new PluginExecutionException("The detected Podman client version number is not supported:" + podmanVersion.trim() + ". Podman version must be 4.4.4 or higher.");
```
cherylking, line 102: Use the constant in the exception for the minimum version in case it changes in the future. Then we only have to change in one place.

src/main/java/io/openliberty/tools/common/plugins/util/AbstractContainerSupportUtil.java
```diff
@@ -21,21 +45,81 @@ public abstract class AbstractContainerSupportUtil {
...
+    protected String getContainerCommandPrefix() throws PluginExecutionException {
+        if (!checkedContainerType) {
+            checkDockerVersion();
+        }
+
+        if (!isDocker) {
+            return CONTAINER_PODMAN_PREFIX;
+        }
+
+        return CONTAINER_DOCKER_PREFIX;
```
cherylking, line 63: Can simplify lines 59 - 63:

`return !isDocker ? CONTAINER_PODMAN_PREFIX : CONTAINER_DOCKER_PREFIX;`

src/main/java/io/openliberty/tools/common/plugins/util/AbstractContainerSupportUtil.java
```diff
@@ -48,22 +132,26 @@ protected String execDockerCmd(String command, int timeout, boolean throwExcepti
...
-            throw new RuntimeException(e.getMessage());
+            // Logging IOExceptions in info stream. This is thrown if Docker or Podman are not installed on the system.
+            info("An error occurred while running a container command: " + e.getMessage());
+            info("This message will occur when Docker or Podman are not installed.");
         }
         return result;
     }
 
+    protected String execContainerCmd(String command, int timeout) {
+        return execContainerCmd(command, timeout, false);
```
cherylking, line 152: Do you really want the boolean to be `false` here? If so, I think the method name should be different...maybe `execContainerCmdSuppressError`.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -120,6 +122,8 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
     private static final String DEVMODE_IMAGE_SUFFIX = "-dev-mode";
     public static final String SKIP_BETA_INSTALL_WARNING = "skipBetaInstallFeatureWarning";
     public static final String DEVC_HIDDEN_FOLDER = ".libertyDevc";
+    public static final String DEVC_DOCKER_PREFIX = "docker ";
```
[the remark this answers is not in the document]
cherylking: Actually, are these even used here? I thought the `getContainerCommandPrefix()` method was used to get the container type everywhere now?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -803,6 +815,8 @@ public void run() {
                             error(e2.getMessage());
                             serverThreadException = e2;
                         }
+                    } catch (PluginExecutionException pe) {
+                        serverThreadException = pe;
```
cherylking, line 819: Should we log this error?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1503,30 +1501,26 @@ private String[] getCommandTokens(String command) {
...
             error("Error stopping container: " + r.getMessage());
             throw r;
         } finally {
-            dockerRunProcess = null;
+            containerRunProcess = null;
         }
     }
 
-    private String execDockerCmd(String command, int timeout) {
-        return execDockerCmd(command, timeout, true);
```
cherylking, line 1527: See this used to pass `true` for the boolean that controls whether exceptions get thrown, but when you moved this method to `AbstractContainerSupportUtil` you changed it to `false`.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1602,31 +1596,39 @@ private String getContainerCommand() {
         }
 
         // mount potential directories containing .war.xml from devc specific folder - override /config/apps and /config/dropins
-        command.append(" -v " + buildDirectory + "/" + DEVC_HIDDEN_FOLDER + "/apps:/config/apps");
-        command.append(" -v " + buildDirectory + "/" + DEVC_HIDDEN_FOLDER + "/dropins:/config/dropins");
+        File tempApps = new File(buildDirectory, DEVC_HIDDEN_FOLDER + "/apps");
+        File tempDropins = new File(buildDirectory, DEVC_HIDDEN_FOLDER + "/dropins");
+        if (!tempApps.exists()) {tempApps.mkdirs();}
```
cherylking, line 1601: So the code I added for this was in method `startContainer()`. Should we be doing this here and there? If it legitimately needs to be done in both places, maybe we should pull out into a separate method to check for the existence of all three directories and create them if needed (apps, dropins, logs).

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
mattbsox: Seemed to work fine without these additions when I did some testing. I'll remove the checks here.

cherylking (approved):

cherylking:
@mattbsox Just realized that there was no update to the `DevUtil.writeDevcMetadata()` method. Can you add the code to include which type of container it is...docker or podman?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1336,24 +1335,24 @@ private void startContainer() {
...
         } catch (InterruptedException e) {
             error("Thread was interrupted while starting the container: " + e.getMessage());
         } catch (RuntimeException r) {
             try {
-                // remove container in case of an error trying to run the container because the docker run --rm will not rm the container
-                String dockerRmCmd = "docker container rm " + containerName;
-                execDockerCmd(dockerRmCmd, DOCKER_TIMEOUT);
+                // remove container in case of an error trying to run the container because the container run --rm will not rm the container
+                String containerRmCmd = "container rm " + containerName;
+                execContainerCmdWithPrefix(containerRmCmd, DOCKER_TIMEOUT);
```
cherylking, line 1352: I should have caught this before. Make `CONTAINER_TIMEOUT` public in `AbstractContainerSupportUtil` and use it throughout this class. Get rid of `DOCKER_TIMEOUT`.

cherylking (changes requested):

cherylking (approved):

PR #414, opened 2023-09-15
Enable signature verification for feature install 

jjiwooLim:
In addition to https://github.com/OpenLiberty/open-liberty/issues/17220.
FeatureUtility now verifies downloaded features from Maven Central. Enabling the same functionality for Liberty Maven Plugin's `install-feature` goal and Liberty Gradles Plugin's `installFeature` task.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
@@ -725,7 +725,7 @@ public abstract class InstallFeatureUtil extends ServerFeatureUtil {
             }
             
 
-            debug("Installing features: " + featuresToInstall);
+            info("Installing features: " + featuresToInstall);
             StringBuilder installedFeaturesBuilder = new StringBuilder();
             Collection<String> actionReturnResult = new ArrayList<String>();
             for (File esaFile : artifacts) {
@@ -1163,7 +1163,7 @@ public abstract class InstallFeatureUtil extends ServerFeatureUtil {
 
         String featureUtilityCommand = getContainerCommandPrefix() + " exec -e FEATURE_LOCAL_REPO=/devmode-maven-cache " + containerName + " featureUtility installFeature " + featureList;
         if (acceptLicense) {
-            featureUtilityCommand += "--acceptLicense";
+            featureUtilityCommand += " --acceptLicense";
         }
         
 
--- a/src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
@@ -1150,7 +1150,7 @@ public abstract class InstallFeatureUtil extends ServerFeatureUtil {
         }
     }
 
-    private void installFeaturesOnContainer(List<String> features, boolean acceptLicense) throws PluginExecutionException {
+    private void installFeaturesOnContainer(List<String> features, boolean acceptLicense, VerifyOption verifyOption) throws PluginExecutionException {
         if (features == null || features.isEmpty()) {
             debug("Skipping installing features on container " + containerName + " since no features were specified.");
             return;
@@ -1173,7 +1173,7 @@ public abstract class InstallFeatureUtil extends ServerFeatureUtil {
             featureUtilityCommand += "--verify=" + verifyOption.name();
         }
         
-        String cmdResult = execDockerCmd(featureUtilityCommand, 600, false);
+        String cmdResult = execContainerCmd(featureUtilityCommand, 600, false);
         if (cmdResult.contains(" RC=")) { // This piece of the string is added in execDockerCmd if there is an error
             if (cmdResult.contains("CWWKF1250I")) {
                 // The features are already installed message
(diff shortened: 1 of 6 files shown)


cherylking:
We plan on doing a release by the end of September. Does this need to go into that? This requires changes in ci.maven and ci.gradle too right? And new tests?

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -126,7 +140,7 @@ public abstract class InstallFeatureUtil extends ServerFeatureUtil {
      * @throws PluginExecutionException If properties files cannot be found in the
      *                                  installDirectory/lib/versions
      */
-    public InstallFeatureUtil(File installDirectory, File buildDirectory, String from, String to, Set<String> pluginListedEsas, List<ProductProperties> propertiesList, String openLibertyVersion, String containerName, List<String> additionalJsons) throws PluginScenarioException, PluginExecutionException {
+    public InstallFeatureUtil(File installDirectory, File buildDirectory, String from, String to, Set<String> pluginListedEsas, List<ProductProperties> propertiesList, String openLibertyVersion, String containerName, List<String> additionalJsons, String verifyValue, Collection<Map<String, String>> keyMap) throws PluginScenarioException, PluginExecutionException {
```
cherylking, line 143: Just a thought...if you also provided a constructor without these two parameters and defaulted the keyMap to null and the verifyValue to `skip`....then this PR could get merged without affecting either maven or gradle. The PRs for maven/gradle that utilize the new constructor with these parameters could go in later once everything is ready and tests exist.

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -288,6 +304,21 @@ private File loadInstallJarFile(File installDirectory) {
...
+     * 
+     * @param esa    	 The ESA file for the signature
+     * @param groupId    The group ID
+     * @param artifactId The artifact ID
+     * @param type       The type e.g. esa
+     * @param version    The version
+     * @return The file corresponding to the downloaded artifact
+     * @throws PluginExecutionException If the artifact could not be downloaded
+     */
+    public abstract File downloadSignature(File esa, String groupId, String artifactId, String type, String version)
```
cherylking, line 320: This new abstract method will require the changes to maven/gradle. So ignore my other comment about the constructor.

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -458,7 +489,21 @@ private File downloadEsaArtifact(String mavenCoordinates) throws PluginExecution
...
         String version = mavenCoordinateArray[2];
-        return downloadArtifact(groupId, artifactId, "esa", version);
+        File downloadedEsa = downloadArtifact(groupId, artifactId, "esa", version);
+        if(this.verifyOption != VerifyOption.skip) {
+            //download signature file for this esa
+            try {
+        	downloadSignature(downloadedEsa, groupId, artifactId, "esa.asc", version);
+            }catch(PluginExecutionException e) {
+        	if(this.verifyOption == VerifyOption.all) {
+        	    throw e;
```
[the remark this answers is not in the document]
jjiwooLim: `enforce` verifies all Liberty features. 
`all` verifies both Liberty and user featuers.
`warn` verifies both Liberty and user features, but if there's signature verification failure, it will result in warning message. 

At this point, we don't know if the download failed for Liberty and or user feature, so we don't throw exception for `enforce`.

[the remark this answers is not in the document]
cherylking: In that case, probably would be good to add some comments to the code explaining this.

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -726,10 +725,9 @@ public void installFeatures(boolean isAcceptLicense, List<String> featuresList)
             }
             
 
-            info("Installing features: " + featuresToInstall);
+            debug("Installing features: " + featuresToInstall);
```
cherylking, line 728: This one should stay `info` as it was before.

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -1170,7 +1168,7 @@ private void installFeaturesOnContainer(List<String> features, boolean acceptLic
         
 
         if(verifyOption != null) {
-            featureUtilityCommand += "--verify=" + verifyOption.name();
+            featureUtilityCommand += " --verify=" + verifyOption.name();
```
cherylking, line 1171: Can you please fix the line above to add a space before `--acceptLicense`? I know you didn't add that code, but your new option here made me think about concatenating options and there needing to be spaces between them.

cherylking (changes requested):
See two minor change requests. Thanks.

cherylking (approved):

PR #422, opened 2023-10-18
Test with Java 21

mattbsox:


src/main/java/io/openliberty/tools/common/plugins/util/AbstractContainerSupportUtil.java
```diff
@@ -51,6 +51,11 @@ public abstract class AbstractContainerSupportUtil {
      */
     public abstract void info(String msg);
 
+    public void setIsDocker(boolean isDocker) {
+        this.isDocker = isDocker;
+        this.checkedContainerType = true;
+    }
```
cherylking, line 57: Shouldn't the `checkPodmanVersion` method be called here if `isDocker` is false? Could call `checkDockerVersion` if `isDocker` parameter is true too.

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/AbstractContainerSupportUtil.java
```diff
@@ -96,8 +110,11 @@ private boolean checkPodmanVersion() throws PluginExecutionException  {
         if (VersionUtility.compareArtifactVersion(podmanVersion, MIN_PODMAN_VERSION, false) < 0) {
             throw new PluginExecutionException("The detected Podman client version number is not supported:" + podmanVersion.trim() + ". Podman version must be " + MIN_PODMAN_VERSION + " or higher.");
         }
-        isDocker = false;
-        checkedContainerType = true;
+
+        if (!checkedContainerType) {
+            isDocker = false;
```
cherylking, line 115: I feel like setting `isDocker` to false might be needed even when `checkedContainerType` is true. In the case that `setIsDocker` is called with the parameter set to `true` and Docker is not installed, the code will call `checkPodmanVersion` as the fallback.

mattbsox: Should we skip the Podman check if `checkedContainerType` is true? If there's no Docker, but a valid Podman version, do we want to use Podman even though it was set to use Docker?

PR #425, opened 2023-10-24
Support directory as valid location to include in server configuration

evie-lau:
Fixes OpenLiberty/ci.maven#1720

When a directory is specified in `<include>`, parse all direct children xml files alphabetically.
Added tests for directory include support.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
@@ -444,10 +444,7 @@ public class ServerConfigDocument {
         } else if (loc.startsWith("file:")) {
             if (isValidURL(loc)) {
                 locFile = new File(loc);
-                if (locFile.exists()) {
-                    doc = parseDocument(locFile);
-                    docs.add(doc);
-                }
+                parseDocumentFromFile(locFile, docs);
             }
         } else if (loc.startsWith("ftp:")) {
             // TODO handle ftp protocol
@@ -456,15 +453,7 @@ public class ServerConfigDocument {
 
             // check if absolute file
             if (locFile.isAbsolute()) {
-                if (locFile.exists()) {
-                    if (locFile.isFile()) {
-                        doc = parseDocument(locFile);
-                        docs.add(doc);
-                    }
-                    if (locFile.isDirectory()) {
-                        parseDocumentsInDirectory(locFile, docs);
-                    }
-                }
+                parseDocumentFromFile(locFile, docs);
             } else {
                 // check configDirectory first if exists
                 if (configDirectory != null && configDirectory.exists()) {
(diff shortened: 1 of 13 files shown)


src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -432,14 +439,12 @@ private static Document getIncludeDoc(String loc) throws IOException, SAXExcepti
...
+                docs.add(doc);
             }
         } else if (loc.startsWith("file:")) {
             if (isValidURL(loc)) {
                 locFile = new File(loc);
-                if (locFile.exists()) {
-                    InputStream inputStream = new FileInputStream(locFile.getCanonicalPath());
-                    doc = parseDocument(inputStream);
-                }
+                parseDocumentFromFile(locFile, docs);
```
evie-lau, line 447: Do we need to consider directory case for URL?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -461,13 +463,58 @@ private static Document getIncludeDoc(String loc) throws IOException, SAXExcepti
...
+                } catch (Exception e) {
+                    e.printStackTrace();
+                }
+            });
+    }
+
+    private static Document parseDocument(File file) throws FileNotFoundException, IOException, SAXException {
+        InputStream is = new FileInputStream(file.getCanonicalPath());
+        Document doc = parseDocument(is);
+        is.close();
```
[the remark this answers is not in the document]
evie-lau: I just realized the `parseDocument(InputStream ...)` method you wrote before has a comment saying it auto-closes the InputStream. I'll just simplify this method.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
cherylking: What does the documentation say? What does Liberty do?

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -559,17 +606,19 @@ private static void parseIncludeVariables(Document doc) throws XPathExpressionEx
...
+
+                ArrayList<Document> inclDocs = getIncludeDoc(includeFileName);
+
+                for (Document inclDoc : inclDocs) {
+                    if (inclDoc != null) {
+                        parseVariablesForBothValues(inclDoc);
+                        // handle nested include elements
+                        parseIncludeVariables(inclDoc);
+                    } else {
+                        log.warn("Unable to parse include file "+includeFileName+". Skipping the included file during application location processing.");
```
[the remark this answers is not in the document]
evie-lau: It seems we no longer return `null` for Documents now, and are already properly logged in other methods when initially parsing the Document.
Will remove this redundant/misleading log statement, and possibly the whole if/else statement.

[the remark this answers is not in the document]
cherylking: The `getIncludeDoc` method used to return null for  the single file. Now I presume it could return an empty list? Maybe check for that and log it. But make sure that proper logging is done when a file within a dir is not able to be found/processed for some reason.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
evie-lau: The documentation is lacking here as well. 

From inspecting Liberty code, it seems the include `location` is resolved before checking if it is a directory, so it should also handle URLs. Posed this follow-up question in slack for confirmation.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -559,17 +616,17 @@ private static void parseIncludeVariables(Document doc) throws XPathExpressionEx
...
-                    parseVariablesForBothValues(docIncl);
-                    // handle nested include elements
-                    parseIncludeVariables(docIncl);
-                } else {
-                    log.warn("Unable to parse include file "+includeFileName+". Skipping the included file during application location processing.");
+
+                ArrayList<Document> inclDocs = getIncludeDoc(includeFileName);
+
+                for (Document inclDoc : inclDocs) {
+                    if (inclDoc != null) {
```
cherylking, line 625: inclDoc cannot be null. Should probably check for an empty list and log a warning.

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
evie-lau: the `getIncludeDoc()` method logs a warning before returning the empty list.
Removing null check here

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
cherylking: The answer from Liberty dev team is that it does not support specifying a directory via URL in the `location` attribute for the `include` element.

cherylking (approved):

PR #426, opened 2023-10-26
Modify behavior for include directory to match Liberty

evie-lau:
Match Liberty behavior where <include> directory has to end in File.separator
> Note: Both Liberty and our test code convert Windows `\` into `/`, so all the logic uses `/`

Note implemented behavior in https://github.com/OpenLiberty/ci.common/pull/426#discussion_r1373606247

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
@@ -450,11 +450,8 @@ public class ServerConfigDocument {
             // TODO handle ftp protocol
         } else {
             locFile = new File(loc);
-
             // check if absolute file
-            if (locFile.isAbsolute()) {
-                parseDocumentFromFileOrDirectory(locFile, docs);
-            } else {
+            if (!locFile.isAbsolute()) {
                 // check configDirectory first if exists
                 if (configDirectory != null && configDirectory.exists()) {
                     locFile = new File(configDirectory, loc);
@@ -463,8 +460,8 @@ public class ServerConfigDocument {
                 if (locFile == null || !locFile.exists()) {
                     locFile = new File(getServerXML().getParentFile(), loc);
                 }
-                parseDocumentFromFileOrDirectory(locFile, docs);
             }
+            parseDocumentFromFileOrDirectory(locFile, loc, docs);
         }
 
         if (docs.isEmpty()) {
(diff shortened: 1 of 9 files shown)


src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -475,24 +472,36 @@ private static ArrayList<Document> getIncludeDocs(String loc) throws IOException
...
-            doc = parseDocument(file);
-            docs.add(doc);
+        // If file mismatches Liberty definition of directory
+        if (f.isFile() && isLibertyDirectory) {
+            log.error("Path specified a directory, but resource exists as a file (path=" + loc + ")");
+            return;
+        } else if (f.isDirectory() && !isLibertyDirectory) {
+            log.error("Path specified a file, but resource exists as a directory (path=" + loc + ")");
+            return;
+        }
```
evie-lau, line 497: Note: Liberty will throw an Exception and stop processing if there is a file/directory type mismatch. But it is different from the behavior for other `<include>` errors, where it continues processing the server config xml.

They are investigating the behavior difference. Regardless, I think for our plugins it makes sense to continue processing (with an error log) instead of throwing an Exception.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -475,24 +472,36 @@ private static ArrayList<Document> getIncludeDocs(String loc) throws IOException
 
     /**
      * Parses file or directory for all xml documents, and adds to ArrayList<Document>
-     * @param file - file or directory to parse documents from
+     * @param f - file or directory to parse documents from
+     * @param isLibertyDirectory - indicates if directory. Liberty bases this off of the presence of trailing File.separator
```
cherylking, line 476: This javadoc does not match the parameter being passed - `loc`

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -475,24 +472,36 @@ private static ArrayList<Document> getIncludeDocs(String loc) throws IOException
...
      * @throws FileNotFoundException
      * @throws IOException
      * @throws SAXException
      */
-    private static void parseDocumentFromFileOrDirectory(File file, ArrayList<Document> docs) throws FileNotFoundException, IOException, SAXException {
+    private static void parseDocumentFromFileOrDirectory(File f, String loc, ArrayList<Document> docs) throws FileNotFoundException, IOException, SAXException {
         Document doc = null;
-        if (file == null || !file.exists()) {
-            log.warn("Unable to parse from file: " + file.getCanonicalPath());
+        boolean isLibertyDirectory = loc.endsWith("/"); // Liberty uses this to determine if directory
```
cherylking, line 484: I thought you said that Liberty would change `\` to `/` when processing the location attribute? Don't we need to handle both?

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -475,24 +472,36 @@ private static ArrayList<Document> getIncludeDocs(String loc) throws IOException
...
+            log.error("Path specified a file, but resource exists as a directory (path=" + loc + ")");
+            return;
+        }
+
+        if (f.isDirectory()) {
+            parseDocumentsInDirectory(f, docs);
         }
-        if (file.isDirectory()) {
-            parseDocumentsInDirectory(file, docs);
+        if (f.isFile()) {
```
cherylking, line 502: Do we need to check `isFile()` at this point? If it is not a directory, and it exists, doesn't it have to be a file?

src/test/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtilGetServerFeaturesTest.java
```diff
@@ -598,7 +598,8 @@ private void replaceIncludeLocation(String includeLocation) throws Exception {
      */
     @Test
     public void testIncludeDir() throws Exception {
-        replaceIncludeDir("includeDir");
+        // Note: Both the product code and test code end up converting Windows \ into /
+        replaceIncludeLocation("includeDir/");
```
cherylking, line 602: Should you test without the `/` and prove it doesn't get processed (and the error is logged)?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
cherylking: Is it because the value passed in was already converted? If so, add comment.

evie-lau: `VariableUtility.resolveVariables(...)` already converts everything to `/` earlier on. I added a comment in `ServerFeatureUtil` but forgot to comment in this file/method.

src/test/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtilGetServerFeaturesTest.java
evie-lau: Will add a test for this failing case

cherylking (approved):

PR #427, opened 2023-10-31
Add additional devc properties to metadata file

mattbsox:


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -5701,13 +5701,22 @@ public void writeDevcMetadata(boolean alive) {
...
+            writeElement(metadataWriter, "imageName", imageName);
+            if (containerfile != null) {
+                writeElement(metadataWriter, "containerfile", containerfile.getCanonicalPath());
+            }
+            if (containerBuildContext != null) {
+                writeElement(metadataWriter, "containerBuildContext", containerBuildContext.getCanonicalPath());
+            }
             writeElement(metadataWriter, "containerAlive", String.valueOf(alive));
+            writeElement(metadataWriter, "containerBuildTimeout", Integer.toString(containerBuildTimeout));
+            writeElement(metadataWriter, "containerRunOpts", containerRunOpts);
```
cherylking, line 5713: Are we sure `containerRunOpts` will not be null? What about from LGP?

mattbsox: We skip writing the value for the elements if the property is null. The element will be in the file, just empty.

mattbsox: The write and null check are here: https://github.com/OpenLiberty/ci.common/blob/e69e5a37508715a3ba91fe504398d5e8f560dcc0/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java#L5723-L5727

cherylking (commented):
Just the one question.

cherylking (approved):

PR #429, opened 2023-11-03
Correctly sort dir files alphabetically to match Liberty behavior

evie-lau:
- Use case-insensitive sort for dir files to match Liberty behavior.
- Fix configDropins processing to sort alphabetically
- Simplified and refactored some code.

Current 23.0.0.10 release has a bug where it uses case-sensitive sort for `<include>` dirs.
Fix in open-liberty went in recently, and will be in the 23.0.0.12 release

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
@@ -22,15 +22,11 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.net.URLConnection;
-import java.nio.file.DirectoryStream;
-import java.nio.file.Files;
-import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Comparator;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Set;
-import java.util.stream.StreamSupport;
 import java.util.Map;
 import java.util.Properties;
 
@@ -43,6 +39,7 @@ import javax.xml.xpath.XPathExpression;
 import javax.xml.xpath.XPathExpressionException;
 import javax.xml.xpath.XPathFactory;
 
+import org.apache.commons.io.comparator.NameFileComparator;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.NodeList;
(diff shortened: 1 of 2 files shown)


cherylking:
Please build/run the tests for LMP and LGP with this branch and confirm none of the tests need updating since it is a change in behavior (even though it was a bug and incorrect behavior before). TIA

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -532,9 +522,15 @@ private static void parseDocumentsInDirectory(File directory, ArrayList<Document
      * @throws IOException
      * @throws SAXException
      */
-    private static Document parseDocument(File file) throws FileNotFoundException, IOException, SAXException {
-        InputStream is = new FileInputStream(file.getCanonicalPath());
-        return parseDocument(is);
+    private static Document parseDocument(File file) throws FileNotFoundException, IOException {
+        try (FileInputStream is = new FileInputStream(file)) {
```
cherylking, line 526: The `parseDocument(InputStream in)` method already closes the input stream. Is it going to cause a problem that both this method and that one will try to close the stream? Or does it handle that fine?

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -487,18 +484,10 @@ private ArrayList<File> parseIncludeFileOrDirectory(String includeFileName, File
...
-                    .forEach(p -> { 
-                        try {
-                            includeFiles.add(p.toFile());
-                        } catch (Exception e) {
-                            debug("Failed to resolve file from path: " + p);
-                        }
-                    });
-            } catch (IOException e) {
-                debug("Unable to open include directory: " + includeFileName);
+            File[] files = includeFile.listFiles();
```
cherylking, line 487: Shouldn't we only consider `.xml` files like the previous logic?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
evie-lau: Will have to do some research on this.
It seems there's a difference between manually calling `close()` where I think it no-ops if already closed, vs auto-closing might have a different behavior.

evie-lau:
LMP/LGP builds passed with these changes included.
Used the same build commands as what runs in Github Actions
```
mvn -V verify --batch-mode --no-transfer-progress --errors -DtrimStackTrace=false -Ponline-its -D"invoker.streamLogsOnFailures"=true -Druntime=ol -DruntimeVersion=23.0.0.10

gradle clean install check -P"test.exclude"="**/TestSpringBootApplication30*,**/DevContainerTest*" -Druntime=ol -DruntimeVersion=23.0.0.10 --stacktrace --i
```

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
evie-lau: The close() on `FileInputStream` does nothing if it's already closed. And `InputStream` close() will also effectively do nothing if called twice.

To be safe, I also refactored some code that called these methods to try to prevent resource leaks from unclosed streams.

cherylking (approved):

PR #430, opened 2023-11-10
Update devc metadata file write behavior for container build and start

mattbsox:
Fixed writers not closing fully in `writeDevcMetadata()` and changed `execContainerCmdAndLog()` to only write devc metadata file on container start/stop commands.

Fixes: https://github.com/OpenLiberty/ci.maven/issues/1746

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -80,6 +80,8 @@ import javax.xml.parsers.ParserConfigurationException;
 
 import com.sun.nio.file.SensitivityWatchEventModifier;
 
+import io.openliberty.tools.ant.ServerTask;
+
 import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
@@ -95,8 +97,6 @@ import org.w3c.dom.Document;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
-
-import io.openliberty.tools.ant.ServerTask;
 
 /**
  * Utility class for dev mode.
@@ -1280,7 +1280,9 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                 info("The RUN features.sh command is detected in the Containerfile and extra time may be necessary when installing features.");
             }
             long startTime = System.currentTimeMillis();
-            execContainerCmdAndLog(getRunProcess(buildCmd), containerBuildTimeout);
+
+            execContainerCmdAndLog(getRunProcess(buildCmd), containerBuildTimeout, false);
+
             checkContainerBuildTime(startTime, buildContext);
             info("Completed building container image.");
         } catch (IllegalThreadStateException  e) {
@@ -1338,7 +1340,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
             String startContainerCommand = getContainerCommand();
             info(startContainerCommand);
             containerRunProcess = getRunProcess(startContainerCommand);
-            execContainerCmdAndLog(containerRunProcess, 0);
+            execContainerCmdAndLog(containerRunProcess, 0, true);
         } catch (IOException e) {
             error("Error starting container: " + e.getMessage());
             throw new RuntimeException(e);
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1280,7 +1282,12 @@ private void buildContainerImage(File tempContainerfile, File userContainerfile,
                 info("The RUN features.sh command is detected in the Containerfile and extra time may be necessary when installing features.");
             }
             long startTime = System.currentTimeMillis();
-            execContainerCmdAndLog(getRunProcess(buildCmd), containerBuildTimeout);
+            //Run build command with an Ant task if using Podman on Windows, see https://github.com/OpenLiberty/ci.maven/issues/1746
+            if (!isDocker && OSUtil.isWindows()) {
+                containerAntRun(buildCmd);
```
cherylking, line 1287: So this is not considering the `containerBuildTimeout`? Should it?

src/main/java/io/openliberty/tools/common/plugins/util/AntTaskFactory.java
```diff
@@ -0,0 +1,94 @@
...
+import java.io.PrintStream;
+import java.util.Objects;
+
+import org.apache.tools.ant.DefaultLogger;
+import org.apache.tools.ant.Project;
+import org.apache.tools.ant.Task;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AntTaskFactory {
```
cherylking, line 28: Can we use this in ci.maven now and delete the similar class from that repo?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -1280,7 +1282,12 @@ private void buildContainerImage(File tempContainerfile, File userContainerfile,
...
+            } else {
+                execContainerCmdAndLog(getRunProcess(buildCmd), containerBuildTimeout);
```
cherylking, line 1289: There are some things that this method does we need to account for in this new `containerAntRun` method. For example, the `writeDevcMetadata(true)` is called. There is also some logic around stopping dev mode if the container does not start successfully. And there are some threads to capturing the input and error stream.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -5696,27 +5703,35 @@ private Set<String> getGeneratedFeatures() {
...
-            }
-            writeElement(metadataWriter, "containerAlive", String.valueOf(alive));
-            writeElement(metadataWriter, "containerBuildTimeout", Integer.toString(containerBuildTimeout));
-            writeElement(metadataWriter, "containerRunOpts", containerRunOpts);
-            metadataWriter.writeEndElement();
-            metadataWriter.writeEndDocument();
-            metadataWriter.flush();
-            metadataWriter.close();
+            FileWriter metaFileWriter = new FileWriter(metaFile);
+            XMLStreamWriter metadataWriter = XMLOutputFactory.newInstance().createXMLStreamWriter(metaFileWriter) ;
```
cherylking, line 5707: There is the slight possibility that this could leave open the `FileWriter` if there is an exception when creating the `XMLStreamWriter`. Should just declare these vars and set to null here, then create inside the `try` that has the `finally`. In the `finally`, only call flush/close if they are not null.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -5696,27 +5703,35 @@ private Set<String> getGeneratedFeatures() {
...
+                warn("Failed to write metadata.\n" + e.getMessage());
+            } finally {
+                metadataWriter.flush();
+                metadataWriter.close();
+                metaFileWriter.flush();
+                metaFileWriter.close();
+            }
         } catch (Exception e) {
-            warn("Failed to write metadata.\n" + e.getMessage());
+            warn("Failed to create metadata writer.\n" + e.getMessage());
```
cherylking, line 5734: "Failed to write metadata due to error creating writer.\n" + e.getMessage()

cherylking (changes requested):
Should edit the summary in the PR to reflect the current changes.

cherylking (approved):

PR #433, opened 2023-11-28
Add config to not fail on javadoc errors

cherylking:


mattbsox (approved):
I think this is good assuming everything works with the parent pom.

PR #439, opened 2024-02-05
Make some methods public, add simple init for logger and props

evie-lau:


src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -180,6 +180,12 @@ public static ServerConfigDocument getInstance(CommonLoggerI log, File serverXML
         return instance;
     }
 
+    public static void setLogger(CommonLoggerI log) {
+        ServerConfigDocument.log = log;
```
cherylking, line 184: The `log` is now non-static as are the other methods. Need to add a new constructor that takes the minimum parms needed, and instantiate this class to call the other methods.

cherylking (changes requested):

cherylking (approved):

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -121,6 +121,12 @@ public ServerConfigDocument(CommonLoggerI log, File serverXML, File configDir, F
         initializeAppsLocation(log, serverXML, configDir, bootstrapFile, bootstrapProp, serverEnvFile, giveConfigDirPrecedence, libertyDirPropertyFiles);
     }
 
+    public ServerConfigDocument(CommonLoggerI log) {
+        this.log = log;
+        props = new Properties();
+        defaultProps = new Properties();
```
cherylking, line 127: I think you also need to initialize `libertyDirectoryPropertyToFile`. It is used in `VariableUtility.getPropertyValue` by one of the methods you made public below.

cherylking (changes requested):

PR #441, opened 2024-02-22
Build docker commands better

cherylking:
For ci.maven issue https://github.com/OpenLiberty/ci.maven/issues/1797

Handle building the docker commands for devc better. By adding each command option/argument separately, the ProcessBuilder can handle paths with spaces. Before we were building one big String, and then turning it into a String[] using a tokenizer that used spaces (among other things like carriage returns, line feeds, etc) to separate the String into command arguments. This caused some arguments to get split into multiple, which was obviously not correct.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -1267,21 +1267,30 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
             // Name rules: may contain lowercase letters, digits and a period, one or two underscores, or one or more dashes. Cannot start with dash.
             imageName = imageName.replaceAll("[^a-zA-Z0-9]", "-").replaceAll("^[\\-]+", "").toLowerCase();
 
-            StringBuilder sb = new StringBuilder();
-            sb.append(getContainerCommandPrefix());
-            sb.append(" build ");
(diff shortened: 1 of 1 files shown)


cherylking:
With this PR, I used the docker getting started guide and ran `mvn liberty:devc` successfully. The `docker build` command looked like this (took the String[] and joined it with a single space as delimiter to log the command). You don't see the quoting/escaping that gets done by ProcessBuilder, but the fact the command runs successfully proves it is handling the path with spaces correctly.

```
[INFO] Building container image...
[INFO] Container command: docker build --pull -f /Users/cherylking/gitForks/guide-docker/finish with space/target/.libertyDevc/tempContainerfile5250737592129611435 -t guide-docker-dev-mode /Users/cherylking/gitForks/guide-docker/finish with space
```

Here is an image proving devc was successful:

Screenshot 2024-02-22 at 4 22 28 PM

Before this PR, the same devc command resulted in the following error:

```
[ERROR] Failed to execute goal io.openliberty.tools:liberty-maven-plugin:3.10:devc (default-cli) on project guide-docker: Error starting the server in dev mode.: Could not build container image using Containerfile: /Users/cherylking/gitForks/guide-docker/finish with space/Dockerfile. Address the following container build error and then start dev mode again: DEPRECATED: The legacy builder is deprecated and will be removed in a future release. RC=1 -> [Help 1]
```

mattbsox (approved):

Issue #126, opened 2020-01-30
Handle additional conditions for resolving variables


cherylking:
In ServerConfigDocument, there is logic that resolves variables referenced by server.xml for the application location and name. This logic processes all the config files that Liberty looks in, but does not look at system properties or environment variables. The document here (https://openliberty.io/docs/ref/config/) indicates those should be used. In a build environment, perhaps that does not make sense. But we should look into it and ensure this behavior is intentional.

That document also covers additional syntax for resolving variable references that we are not handling. We should add logic for this. Here is an excerpt:

Replace all non-alpha num characters with _

Change all characters to upper case.

If you enter ${my.env.var} in server.xml it will look for environment variables with the following names:

my.env.var

my_env_var

MY_ENV_VAR

cherylking:
This part of the Liberty doc is also new since this code was last updated. Need to handle this as well.

![Screenshot 2023-11-07 at 1 44 10 PM](https://github.com/OpenLiberty/ci.common/assets/29436451/4d273b9c-9859-4d9e-9ced-dc567f02986f)

cherylking:
This has been addressed with the referenced PRs.

PR #442, opened 2024-02-27
Update Variable Processing

dshimo:
#126 

- [x] Check compatibility with Maven and Gradle

Maven: https://github.com/dshimo/ci.maven/actions/runs/8377634883
Gradle: https://github.com/dshimo/ci.gradle/actions/runs/8442718868

### Tests
- [x] var extraction
  - +: server.xml; boostrap.properties; server.env; 
  - -: jvm.options; CLI
- [x] overrides on each level
- [x] overall procedure check
- [x] server.env additional file locations
- [x] jvm.options additional file locations
- [x] environment variable variations lookup
- [x] variables directory
- [x] bootstrap.include (Document [link](https://openliberty.io/docs/latest/reference/config/server-configuration-overview.html#bootstrap-properties))

### Config File Precedence (additional file location support)
- [x] server.env
- [x] jvm.options

### New Features
- [ ] environment variable variations lookup
  - [x] 3 step lookup
  - [ ] lists()
  - [ ] Simple arithmetic
- [x] variables directory
- [x] bootstrap.include


src/test/java/io/openliberty/tools/common/config/ServerConfigDocumentTest.java
```diff
@@ -0,0 +1,218 @@
...
+    //   1. {wlp.install.dir}/etc
+    //   2. {wlp.user.dir}/shared
+    //   3. {server.config.dir}
+    // Table of directories: https://openliberty.io/docs/latest/reference/directory-locations-properties.html
+    @Test
+    public void processServerEnv() throws FileNotFoundException, Exception {
+        File serverDir = new File(RESOURCES_DIR + WLP_USER_DIR + DEFAULT_USER_SERVER_DIR);
+        File specificFile = new File(serverDir, "server2.env");
+        ServerConfigDocument configDocument = new ServerConfigDocument(new TestLogger());
+        configDocument.initializeFields(new TestLogger(), null, serverDir, null);
```
cherylking, line 96: I think you should consider passing in the Map<String,File> for the last parameter on `initializeFields` with correct values set for all the pre-defined Liberty variables. Then the `processServerEnv` would be able to do the right thing for precedence related to the number 1 and 2 locations.

src/test/java/io/openliberty/tools/common/config/ServerConfigDocumentTest.java
```diff
@@ -0,0 +1,218 @@
...
+        configDocument.processServerEnv(specificFile, false);
+        assertEquals("TEST", configDocument.getProperties().getProperty("keystore_password"));
+
+        configDocument = new ServerConfigDocument(new TestLogger());
+        configDocument.initializeFields(new TestLogger(), null, serverDir, null);
+        configDocument.processServerEnv(specificFile, true);
+        assertNotEquals("TEST", configDocument.getProperties().getProperty("keystore_password"));
```
cherylking, line 103: Why is the same file and value being tested twice?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -154,65 +166,31 @@ private DocumentBuilder getDocumentBuilder() {
...
-            // 6. variables defined in server.xml
-            // e.g. <variable name="myVarName" value="myVarValue" />
-            // 7. variables from configDropins/overrides/<file_name>
+            // Server variable precedence in ascending order if defined in multiple locations.
+            //  1. variable default values in the server.xml file
+            //  2. environment variables
+            //  3. bootstrap.properties
+            //  4. Java system properties
+            //  5. Variables loaded from files in the ${server.config.dir}/variables directory or other directories as specified by the VARIABLE_SOURCE_DIRS environment variable
+            //  6. variable values declared in the server.xml file
```
cherylking, line 179: Why have we lost the detail of where configDropins fit?

cherylking: I think there is some confusion between two different lists on this [page](https://openliberty.io/docs/latest/reference/config/server-configuration-overview.html). There is a list for file processing [here](https://openliberty.io/docs/latest/reference/config/server-configuration-overview.html#configuration-files). Then there is a list of variable substitution precedence [here](https://openliberty.io/docs/latest/reference/config/server-configuration-overview.html#variable-substitution). We need the comments here to kind of merge both of those into what the code logic should do.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -238,6 +216,95 @@ private void initializeAppsLocation(CommonLoggerI log, File serverXML, File conf
...
+     *   1. {wlp.user.dir}/
+     *   2. {server.config.dir}/configDropins/defaults/
+     *   3. {server.config.dir}/
+     *   4. {server.config.dir}/configDropins/overrides/
+     * @throws FileNotFoundException
+     * @throws Exception
+     */
+    public void processJvmOptions() throws FileNotFoundException, Exception {
+        final String jvmOptionsString = "jvm.options";
+        parsePropertiesFromFile(new File(libertyDirectoryPropertyToFile.get(ServerFeatureUtil.WLP_USER_DIR), jvmOptionsString));
```
cherylking, line 231: The Liberty doc says the first location should be `{wlp.user.dir}/shared`. So I think you need to tack on `shared/` before `jvm.options`.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -238,6 +216,95 @@ private void initializeAppsLocation(CommonLoggerI log, File serverXML, File conf
...
+        parsePropertiesFromFile(getFileFromConfigDirectory(CONFIGDROPINS_DEFAULT + File.separator + jvmOptionsString));
```
cherylking, line 232: So I started wondering if configDirectory is the one specified in the pom.xml (or the default src/main/liberty/config), or if it should just be `server.config.dir` at this point? I looked at what LMP/LGP are passing in to this class, and I think it is what was configured in the build file. But it seems like we should just be concerned with the config files that got copied into the server itself at this point, in which case we could just use `libertyDirectoryPropertyToFile.get(ServerFeatureUtil.SERVER_CONFIG_DIR)` as the basis for the configDropins. Let me know if you didn't follow all of that.

cherylking: I think this could also solve a problem we have right now that we are not necessarily passing in all the right info for bootstrap props and server env (because of overriding Maven props for example), but if we used what got copied/generated in the server it should all be correct at that point.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -238,6 +216,95 @@ private void initializeAppsLocation(CommonLoggerI log, File serverXML, File conf
...
+     *   3. {server.config.dir}/
+     * @param serverEnvFile
+     * @throws Exception
+     * @throws FileNotFoundException
+     */
+    public void processServerEnv() throws Exception, FileNotFoundException {
+        final String serverEnvString = "server.env";
+        parsePropertiesFromFile(new File(libertyDirectoryPropertyToFile.get(ServerFeatureUtil.WLP_INSTALL_DIR), "etc" + File.separator + serverEnvString));
+        parsePropertiesFromFile(new File(libertyDirectoryPropertyToFile.get(ServerFeatureUtil.WLP_USER_DIR), "shared" + File.separator + serverEnvString));
+        parsePropertiesFromFile(new File(libertyDirectoryPropertyToFile.get(ServerFeatureUtil.SERVER_CONFIG_DIR), serverEnvString));
```
cherylking, line 250: So this is what I suggested for processing jvm.options above. You are loading server.env from SERVER_CONFIG_DIR, not from configDirectory.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -238,6 +216,95 @@ private void initializeAppsLocation(CommonLoggerI log, File serverXML, File conf
...
+    }
+
+    /**
+     * Process bootstrap.properties and boostrap.include 
+     * @param bootstrapProp
+     * @param bootstrapFile - Optional specific file
+     * @throws Exception
+     * @throws FileNotFoundException
+     */
+    public void processBootstrapProperties(Map<String, String> bootstrapProp, File bootstrapFile) throws Exception, FileNotFoundException {
```
cherylking, line 284: Hopefully I am thinking correctly, but given what I was saying earlier I think we do not need to pass in separate props and files...just use whatever got copied/generated for bootstrap.properties in SERVER_CONFIG_DIR. Just need to make sure that `copyConfigFiles` is getting called before `ServerConfigDocument` is used in each case in LMP/LGP. I think it is.

cherylking (changes requested):
I've only looked at code changes at this point, and not the tests.

cherylking (changes requested):
I still need to review the test case and all the test case resources, but I wanted you to have my feedback/questions on the code itself. Sorry it has taken me this long. Crazy busy week.

src/test/java/io/openliberty/tools/common/config/ServerConfigDocumentOverridesTest.java
```diff
@@ -0,0 +1,246 @@
...
+        assertEquals(1, configDocument.getProperties().size());
+        assertEquals("9081", configDocument.getProperties().getProperty("http.port"));
+        
+        // server.xml configDropins precedence
+        File serverConfigDir = SERVER_CONFIG_DIR.toFile();
+        libertyDirPropMap.put(ServerFeatureUtil.SERVER_CONFIG_DIR, serverConfigDir);
+        configDocument = new ServerConfigDocument(new TestLogger(), libertyDirPropMap);
+        doc = configDocument.parseDocument(new File(serverConfigDir, "server.xml"));
+        configDocument.processServerXml(doc);   // Variable resolution warnings can be ignored here
+        assertEquals("1", configDocument.getProperties().getProperty("config.dropins.defaults"));
```
cherylking, line 79: How am I supposed to know if these are correct? It would be nice to have a comment here explaining where each one is defined and overridden so I can tell if what I think should be the value is the actual value.

cherylking (changes requested):
I think the tests are good, but it was hard to know for sure. More comments would be great anywhere that you are verifying precedence (explaining where a particular variable was defined and overridden and the various values associated with it).

src/main/java/io/openliberty/tools/common/plugins/util/VariableUtility.java
```diff
@@ -90,35 +89,58 @@ public static String resolveVariables(CommonLoggerI log, String nodeValue, Colle
...
+        if (value.startsWith("\"") && value.endsWith("\"") && value.length() > 2) {
+            return value.substring(1, value.length() - 1);
+        }
         return value;
     }
-    
+
+    private static String lookupProperty(Properties prop, Properties defaultProps, String propertyName) {
+        if (prop.containsKey(propertyName)) {
+            return stripQuotes(prop.getProperty(propertyName));
```
cherylking, line 139: Just because `prop` contains the key does not mean the value from `getProperty` is not null. I think it would be best to guard against a NPE in `stripQuotes` before calling methods on `value`.

src/test/java/io/openliberty/tools/common/config/ServerConfigDocumentOverridesTest.java
```diff
@@ -0,0 +1,255 @@
...
+
+        // bootstrap.include termination check
+        libertyDirPropMap.put(ServerFeatureUtil.SERVER_CONFIG_DIR, new File(serversDir, "bootstrapOuroboros"));
+        configDocument = new ServerConfigDocument(new TestLogger(), libertyDirPropMap, null);
+        configDocument.processBootstrapProperties();
+    }
+
+    // 4. Java system properties
+    @Test
+    public void jvmOptions() {
```
cherylking, line 182: why?

src/test/java/io/openliberty/tools/common/config/ServerConfigDocumentOverridesTest.java
```diff
@@ -0,0 +1,255 @@
...
+
+        props = configDocument.getProperties();
+        assertEquals("outer_space", props.getProperty("outer.source"));
+        assertEquals("1", props.getProperty("VALUE_1"));
+    }
+    
+    // 7. variables declared on the command line
+    @Test
+    public void CLI() {
+
```
cherylking, line 221: why?

cherylking (changes requested):
just a couple small comments to address

cherylking:
Superseded by #462.

PR #447, opened 2024-07-15
Update messages related to running tests when skipTests is set to true

cherylking:
For issue https://github.com/OpenLiberty/ci.gradle/issues/889

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2529,14 +2529,15 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
 
     private void printTestsMessage(boolean formatForAttention) {
         // setting skipTests to true overrides the setting of hotTests, and prevents any tests from running in dev mode (automatically or on demand)
-        if (!skipTests) {
-             if (hotTests) {
-                String message = "Tests will run automatically when changes are detected. You can also press the Enter key to run tests on demand.";
-                info(formatForAttention ? formatAttentionMessage("Enter - " + message) : message);
-            } else {
-                String message = "run tests on demand, press Enter.";
-                info(formatForAttention ? formatAttentionMessage("Enter - " + message) : "To " + message);
-            }
+        if (skipTests) {
+            return;
+        }
+        if (hotTests) {
+            String message = "Tests will run automatically when changes are detected. You can also press the Enter key to run tests on demand.";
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -691,6 +691,12 @@ public void runTests(boolean waitForApplicationUpdate, int messageOccurrences, T
                     error(e.getMessage(), e);
                 }
             }
+        } else {
```
dshimo, line 694: We can simplify/clarify this method a step by removing the wrapping `if (!forceSkipTests)`. The `else` logic can go into an `if (forceSkipTests)` placed at the beginning that does a single `return` after either info messages.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2522,12 +2528,15 @@ private void printDevModeMessages(boolean inputUnavailable, boolean startup) thr
...
 
     private void printTestsMessage(boolean formatForAttention) {
-        if (hotTests) {
-            String message = "Tests will run automatically when changes are detected. You can also press the Enter key to run tests on demand.";
-            info(formatForAttention ? formatAttentionMessage("Enter - " + message) : message);
-        } else {
-            String message = "run tests on demand, press Enter.";
-            info(formatForAttention ? formatAttentionMessage("Enter - " + message) : "To " + message);
+        // setting skipTests to true overrides the setting of hotTests, and prevents any tests from running in dev mode (automatically or on demand)
+        if (!skipTests) {
```
dshimo, line 2532: In a similar line of thought to manage nested conditions, this could also be expressed as
``` 
if (skipTests) {
    return;
}
```

dshimo (changes requested):
An early return pattern can simplify conditional management/wrapping

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
cherylking: Although this is true, I didn't want to change the existing logic when just adding a new log message.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
cherylking: This one I wouldn't mind changing since it doesn't affect the structure of the rest of the method. I'm not going to run through all the PR testing for LMP/LGP again though if I change this.

dshimo (approved):

PR #450, opened 2024-08-28
Start parsing platforms, along with features in featureManager node

cbridgha:
The parsing logic now collects `<platform>` elements along with `<features>`

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
@@ -393,7 +393,7 @@ public abstract class ServerFeatureUtil extends AbstractContainerSupportUtil imp
                         Element child = (Element) nodes.item(i);
                         if ("featureManager".equals(child.getNodeName())) {
                             if (result == null) {
-                                result = new FeaturesPlatforms(new HashSet<String>(), new HashSet<String>());
+                                result = new FeaturesPlatforms();
                             }
                             FeaturesPlatforms fp = parseFeatureManagerNode(child);
                             result.getFeatures().addAll(fp.getFeatures());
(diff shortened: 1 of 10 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -390,19 +447,48 @@ private Set<String> parseFeatureManagerNode(Element node) {
...
             }
         }
-        return result;
+        NodeList platformElements = node.getElementsByTagName("platform");
+        if (platformElements != null) {
+            for (int j = 0; j < platformElements.getLength(); j++) {
+                String content = platformElements.item(j).getTextContent();
+                if (content != null) {
+                	content = content.trim();
+                	if (content.contains(":")) {
```
cherylking, line 462: Checking for `:` was for handling user features. I don't think that applies to the platform elements.

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -339,9 +393,11 @@ public void error(SAXParseException e) throws SAXException {
                         Element child = (Element) nodes.item(i);
                         if ("featureManager".equals(child.getNodeName())) {
                             if (result == null) {
-                                result = new HashSet<String>();
+                                result = new FeaturesPlatforms(new HashSet<String>(), new HashSet<String>());
```
cherylking, line 396: Why not call the empty constructor here?

cherylking (changes requested):
I presume more changes are coming to use the platforms when installing the features?

src/main/java/io/openliberty/tools/common/plugins/util/ServerFeatureUtil.java
```diff
@@ -390,19 +447,39 @@ private Set<String> parseFeatureManagerNode(Element node) {
...
+            }
+        }
+        return new FeaturesPlatforms(features, platforms);
     }
     
     /**
      * Parse features from an include node.
      * 
-     * @param origResult
+     * @param result2
```
cherylking, line 482: Small thing...this param name doesn't match what is in the method signature

cherylking (approved):
Looks good. Should not merge this PR before getting good builds in ci.maven and ci.gradle pointing to this branch though.

PR #455, opened 2024-09-25
Sast security fix

arunvenmany-ibm:


src/main/java/io/openliberty/tools/common/arquillian/util/HttpPortUtil.java
```diff
@@ -46,16 +47,21 @@ public class HttpPortUtil {
...
+        factory.setFeature("http://apache.org/xml/features/nonvalidating/load-dtd-grammar", false);
+        factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
+        factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+        factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
+        factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
+        factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        factory.setXIncludeAware(false);
+        factory.setNamespaceAware(true);
+        factory.setExpandEntityReferences(false);
+        return factory;
```
cherylking, line 63: Are we not allowed to have a singleton instance of the factory? We were only doing the initialization once before. This can get called many times from maven/gradle plugins. This new code is creating a new one every single time.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -229,20 +239,20 @@ private void initializeAppsLocation(CommonLoggerI log, File serverXML, File conf
...
     }
 
     //Checks for application names in the document. Will add locations without names to a Set
-    private void parseNames(Document doc, String expression) throws XPathExpressionException, IOException, SAXException {
+    private void parseNames(Document doc, XPathExpression expression) throws XPathExpressionException, IOException, SAXException {
         // parse input document
-        XPath xPath = XPathFactory.newInstance().newXPath();
-        NodeList nodeList = (NodeList) xPath.compile(expression).evaluate(doc, XPathConstants.NODESET);
+        NodeList nodeList = (NodeList) expression
+                .evaluate(doc, XPathConstants.NODESET);
```
cherylking, line 255: spacing is weird here

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -511,8 +521,14 @@ private Document parseDocument(URL url) throws IOException, SAXException {
     }
 
     private Document parseDocument(InputStream in) throws SAXException, IOException {
-        try (InputStream ins = in) { // ins will be auto-closed
+        InputStream ins = null;
```
cherylking, line 524: Why are we getting rid of try with resources which is the new better way to ensure streams are closed?

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -52,8 +52,8 @@ public abstract class BinaryScannerUtil {
             "in the application’s API usage: %s. Review and update your application to ensure it is not using conflicting APIs " +
             "from different levels of MicroProfile, Java EE, or Jakarta EE.";
     public static final String BINARY_SCANNER_CONFLICT_MESSAGE4 = "[None available]"; // format should match JVM Set.toString()
-    public static final String BINARY_SCANNER_CONFLICT_MESSAGE5 = "A working set of features could not be generated due to conflicts " + 
-            "in the required features: %s and required levels of MicroProfile: %s, Java EE or Jakarta EE: %s. Review and update your application to ensure it " + 
+    public static final String BINARY_SCANNER_CONFLICT_MESSAGE5 = "A working set of features could not be generated due to conflicts " +
```
cherylking, line 55: Are there any actual changes to this file? or all formatting? If the latter, please revert the changes to this file.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -947,7 +948,7 @@ public void onFileChange(File file) {
             // Parse hostname, http, https ports for integration tests to use
             parseHostNameAndPorts(serverTask, messagesLogFile);
         } catch (IOException e) {
-            throw new PluginExecutionException("An error occurred while starting the server: " + e.getMessage(), e);
+            throw new PluginExecutionException("An error occurred while starting the server: " + e.getMessage());
```
cherylking, line 951: I don't think we should do these changes. There is no security risk here because this is all open source. We need this level of detail to support our customers when they report issues. We can document this reasoning in the spreadsheet for all of this category of issues.

src/main/java/io/openliberty/tools/common/plugins/util/ServerStatusUtil.java
```diff
@@ -53,7 +53,7 @@ public static boolean isServerRunning(File installDirectory, File outputDirector
                         return false;
                     }
                 } catch (Exception e) {
-                    e.printStackTrace();
+                    System.out.println("exception while reading pidfile " + e.getMessage());
```
cherylking, line 56: We should not do `System.out.println`. Can a logger be passed in?

cherylking (changes requested):

PR #462, opened 2024-11-08
Var processing changes for ci gradle and ci maven. 

arunvenmany-ibm:
to be used for LCLS as well


src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -182,18 +151,20 @@ public ServerConfigDocument(CommonLoggerI log, Map<String, File> libertyDirPrope
         locationsAndNames = new HashMap<String, String>();
         props = new Properties();
         defaultProps = new Properties();
-
+        if (originalServerXMLFile != null) {
```
cherylking, line 154: Why are we safeguarding setting this field to null, when it defaults to null anyway. Seems unnecessary.

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -205,6 +176,9 @@ public ServerConfigDocument(CommonLoggerI log, Map<String, File> libertyDirPrope
         props = new Properties();
         if (initProperties != null) props.putAll(initProperties);
         defaultProps = new Properties();
+        if (originalServerXMLFile != null) {
```
cherylking, line 179: Why are we safeguarding setting this field to null, when it defaults to null anyway. Seems unnecessary.

cherylking (changes requested):

cherylking (approved):

PR #466, opened 2024-12-23
Server multi restart issue fix

arunvenmany-ibm:


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -441,17 +441,20 @@ private enum FileTrackMode {
...
-            String packagingType, File buildFile, Map<String, List<String>> parentBuildFiles, boolean generateFeatures,
-            Set<String> compileArtifactPaths, Set<String> testArtifactPaths, List<Path> monitoredWebResourceDirs) {
+                   File configDirectory, File projectDirectory, File multiModuleProjectDirectory, List<File> resourceDirs, boolean changeOnDemandTestsAction,
+                   boolean hotTests, boolean skipTests, boolean skipUTs, boolean skipITs, boolean skipInstallFeature, String applicationId,
+                   long serverStartTimeout, int appStartupTimeout, int appUpdateTimeout, long compileWaitMillis,
+                   boolean libertyDebug, boolean useBuildRecompile, boolean gradle, boolean pollingTest, boolean container,
+                   File containerfile, File containerBuildContext, String containerRunOpts, int containerBuildTimeout,
+                   boolean skipDefaultPorts, JavaCompilerOptions compilerOptions, boolean keepTempContainerfile,
+                   String mavenCacheLocation, List<ProjectModule> upstreamProjects, boolean recompileDependencies,
```
cherylking, line 454: Why not provide a second constructor without the Map since Gradle does not need it? Then when that constructor is used, you could initialize the Map to an empty Map instead of making the Gradle code do that.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4021,15 +4025,27 @@ private void initWatchLoop() throws IOException {
         // initial source and test compile of upstream projects
         if (isMultiModuleProject()) {
             for (ProjectModule project : upstreamProjects) {
-                triggerUpstreamModuleCompile(project, false);
+                // if map is empty, should trigger compile
+                if (projectRecompileMap.isEmpty() || Boolean.TRUE.equals(projectRecompileMap.get(project.getProjectName())) ) {
+                    info("Recompile " + project.getProjectName() + " due to an earlier compilation error");
+                    triggerUpstreamModuleCompile(project, false);
```
cherylking, line 4031: Once this compile is triggered for the project that had TRUE in the projectRecompileMap, does something remove it from the Map or set it to FALSE?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
arunvenmany-ibm: No. we are only using this map at the beginning of initWatchLoop() , just to skip the recompilation there.
Whenever there is file change , processFileChanges() is triggered, which triggers triggerProjectCompile() method separately.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
arunvenmany-ibm: Done

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4021,15 +4049,27 @@ private void initWatchLoop() throws IOException {
         // initial source and test compile of upstream projects
         if (isMultiModuleProject()) {
             for (ProjectModule project : upstreamProjects) {
-                triggerUpstreamModuleCompile(project, false);
+                // if map is empty, should trigger compile
+                if (projectRecompileMap.isEmpty() || Boolean.TRUE.equals(projectRecompileMap.get(project.getProjectName())) ) {
```
cherylking, line 4053: When exactly can the map be empty? 1) When called from Gradle? If so, the `info` message at 4054 may be misleading? 2) When no project was identified as needing to be recompiled in Maven? If so, why would we do a recompile here?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4021,15 +4049,27 @@ private void initWatchLoop() throws IOException {
...
                 // build file tracking of upstream projects
                 lastBuildFileChange.put(project.getBuildFile(), System.currentTimeMillis());
             }
         }
 
         // initial source and test compile
-        triggerMainModuleCompile(false);
-        // build file tracking of main project
+        // if map is empty, should trigger compile
+        if (projectRecompileMap.isEmpty()|| Boolean.TRUE.equals(projectRecompileMap.get(getProjectName())) ) {
```
cherylking, line 4066: Same comment about what an empty projectRecompileMap means and how it should affect the `info` messages.

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
arunvenmany-ibm: Map can be empty when passed from gradle. 
Gradle is already using Gradle task state to check whether a recompile is required. Hence even if we always trigger compile from our end, it just returns task state as EVERYTHING UP-TO-DATE

cherylking: But can't it also be empty coming from Maven if no projects had compile errors?

cherylking: And since the answer for 1 is yes, my comment about the misleading messages applies.

arunvenmany-ibm: Changed to set projectRecompileMap as null for gradle and updated message

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
arunvenmany-ibm: Changed to set projectRecompileMap as null for gradle and updated message

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4021,15 +4050,28 @@ private void initWatchLoop() throws IOException {
         // initial source and test compile of upstream projects
         if (isMultiModuleProject()) {
             for (ProjectModule project : upstreamProjects) {
-                triggerUpstreamModuleCompile(project, false);
+                // for ci.gradle, map will be null always
+                // we can always trigger recompile for gradle, as we gradle is using gradle task state to recompile
+                if (projectRecompileMap == null || Boolean.TRUE.equals(projectRecompileMap.get(project.getProjectName())) ) {
+                    info("Recompile " + project.getProjectName() + " due to an earlier compilation error");
```
cherylking, line 4056: So if we are here for gradle and the map is null, we will print a message saying we are recompiling due to an earlier compilation error. But that may or may not be true, right?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
arunvenmany-ibm: Corrected the if else for gradle and maven

cherylking: Ok, why are we triggering a different method for Gradle here? Should it be `triggerUpstreamModuleCompile`? I think we need a Gradle PR pulling in this branch and running tests to ensure everything is good.

Also, shouldn't we log some message like "Check if recompile for project x needs to be done." or just "Recompile project x." and that will get followed up by the typical Gradle message about a task is UP-TO-DATE?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4065,7 +4067,9 @@ private void initWatchLoop() throws IOException {
 
         // for ci.gradle, map will be null always
         // we can always trigger recompile for gradle, as we gradle is using gradle task state to recompile
-        if (projectRecompileMap == null || Boolean.TRUE.equals(projectRecompileMap.get(getProjectName())) ) {
+        if (projectRecompileMap == null) {
+            triggerMainModuleCompile(false);
```
cherylking, line 4071: I think we need some message here for the gradle scenario.

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
arunvenmany-ibm: We can trigger triggerUpstreamModuleCompile or triggerMainModuleCompile in case of gradle, because when we trigger a compile in gradle for EAR , it actually checks whether it needs to recompile any dependent modules. 
After some thoughts , I have rolled back to triggerMainModuleCompile

triggerUpstreamModuleCompile() or triggerMainModuleCompile() just adds sourceSet to a map, which is used in recompileJava() 

image
the useBuildRecompile() flag is set as true in gradle and set as false in maven

In case of gradle, it triggers compile() method in ci.gradle DevTask class
I think no custom message is needed in case of gradle, since it just trigger gradle task directly, which gives full info

image

cherylking: Think of our log messages as helpful hints when debugging a user issue. In that case, it would be helpful to know why the compile got triggered in Gradle. Are there any other paths into that compile method you show in your last image besides this ci.common method?

If you are going to call a different method here for gradle, please add a comment explaining why.

arunvenmany-ibm: 1.
 a. can you provide some wordings for the ci.gradle related message. Since we are not changing anything affecting ci.gradle in any scenarios, I was not sure to introduce a new message.
 b. compile() is being called for any scenario with RecompileJava() -> called from processUpstreamJavaCompilation() in dev file changes

2. No We will keep the triggerUpstreamModuleCompile() itself

arunvenmany-ibm: Yes, gradle task decide what to be done
I will add a log message saying "Recompile {project_name}

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4021,15 +4050,34 @@ private void initWatchLoop() throws IOException {
...
+            info("Recompile " + getProjectName());
+            triggerMainModuleCompile(false);
+        } else if (Boolean.TRUE.equals(projectRecompileMap.get(getProjectName()))) {
+            info("Recompile " + getProjectName() + " due to an earlier compilation error");
+            triggerMainModuleCompile(false);
+            // build file tracking of main project
+        } else {
+            info("Recompile skipped for " + getProjectName() + " since earlier compilation is successful");
+        }
         lastBuildFileChange.put(buildFile, System.currentTimeMillis());
```
cherylking, line 4081: Sorry I didn't notice this before. Do we really want to update `lastBuildFileChange` if we skip recompile?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
arunvenmany-ibm: Since sourceSet is not being added, I think its okay to keep it 
I have added tests for checking recompile in ci.maven

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -3627,7 +3627,7 @@ private boolean processUpstreamJavaCompilation(List<ProjectModule> upstreamProje
                 boolean compileDownstreamSrc = false;
                 boolean compileDownstreamTest = false;
 
-                boolean pastBuildFileWaitPeriod = System
+                boolean pastBuildFileWaitPeriod = lastBuildFileChange.get(project.getBuildFile()) != null && System
```
cherylking, line 3630: We want this to be set to true when the map returns null.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4055,16 +4055,14 @@ private void initWatchLoop() throws IOException {
...
                 } else if (Boolean.TRUE.equals(projectRecompileMap.get(project.getProjectName()))) {
                     info("Recompile " + project.getProjectName() + " due to an earlier compilation error");
                     triggerUpstreamModuleCompile(project, false);
-                    // build file tracking of upstream projects
-                    lastBuildFileChange.put(project.getBuildFile(), System.currentTimeMillis());
                 } else {
                     info("Recompile skipped for " + project.getProjectName() + " since earlier compilation is successful");
                 }
+                // build file tracking of upstream projects, this update is needed for auto test invocation
+                lastBuildFileChange.put(project.getBuildFile(), System.currentTimeMillis());
```
cherylking, line 4065: This should only be set in the if and else if with the change you made to check for null.

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
cherylking: If you are saying it is needed for our tests to work, then I question how our tests are structured. We should not have code in production that is only for testing purposes.

arunvenmany-ibm: No I am talking about the tests compile invocation, in the processupstreamJavaCompilation(), 
Currently, if there is no change in build file, we are skipping the execution, 
image

But I could see that we have more logic for processing test files

image

cherylking: If you set `pastBuildFileWaitPeriod` to true when `lastBuildFileChange.get()` returns null, it will work correctly.

arunvenmany-ibm: I tried defaulting pastBuildFileWaitPeriod with both true and false, both scenarios are causing many tests to fail

cherylking: It seems to be passing with the latest change.

cherylking (approved):

Issue #464, opened 2024-11-20
Fix check for springBootApplication elements


cherylking:
The check added recently for more than one `springBootApplication` element only considered a single server config file. The check needs to take into account all server config files (via include elements and configDropins). I suggest the code should collect a Map of springBootApplication by id and preserve the correct merge/replace behavior based on where the config resides. Once all the server config files have been processed, check if more than one configured `springBootApplication` location exists and throw the error at that point. We could even include the ids for the elements we found in the message.

PR #467, opened 2025-01-01
springbootapplication node multiple occurrence fix

arunvenmany-ibm:
fix for  #464

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
@@ -512,6 +512,9 @@ public class ServerConfigDocument {
             // add unique values only
             if (!nodeValue.isEmpty()) {
                 if(expression.equals(XPATH_SERVER_SPRINGBOOT_APPLICATION)){
+                    if(springBootAppNodeLocation.isPresent()){
+                        throw new PluginExecutionException("Found multiple springBootApplication elements specified in the server configuration. Only one springBootApplication can be configured per Liberty server.");
+                    }
                     springBootAppNodeLocation = Optional.of(nodeValue);
                 }
                 String resolved = VariableUtility.resolveVariables(log, nodeValue, null, getProperties(), getDefaultProperties(), getLibertyDirPropertyFiles());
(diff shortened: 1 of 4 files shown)


src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -512,6 +512,10 @@ private void parseApplication(Document doc, XPathExpression expression) throws X
             // add unique values only
             if (!nodeValue.isEmpty()) {
                 if(expression.equals(XPATH_SERVER_SPRINGBOOT_APPLICATION)){
+                    // checking whether any springBootAppNodeLocation already configured from other server configuration files
+                    if(springBootAppNodeLocation.isPresent()){
+                        throw new PluginExecutionException("Found multiple springBootApplication elements specified in the server configuration. Only one springBootApplication can be configured per Liberty server.");
+                    }
```
cherylking, line 518: I think we should include the path of the config file where the springBootApplication element was found in the error message. Also include in above message at line 508, and add an `else` to this `if` to print a log.debug statement with the config file location of the first springBootApplication element found.

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -533,6 +528,20 @@ private void parseApplication(Document doc, XPathExpression expression) throws X
         }
     }
 
+    private void checkForSpringBootApplicationNode(Document doc, XPathExpression expression, String nodeValue) throws PluginExecutionException {
+        if(expression.equals(XPATH_SERVER_SPRINGBOOT_APPLICATION)){
+            // checking whether any springBootAppNodeLocation already configured from other server configuration files
+            if(springBootAppNodeLocation.isPresent() && springBootAppNodeDocumentURI.isPresent()){
+                throw new PluginExecutionException(String.format("SpringBootApplication elements are specified in multiple server configuration files [%s, %s]. Only one springBootApplication can be configured per Liberty server.", springBootAppNodeDocumentURI.get(), doc.getDocumentURI()));
```
cherylking, line 535: The element name should not be capitalized. Perhaps reword so it doesn't start the sentence. 
"Found multiple springBootApplication elements specified in the server configuration in files [%s, %s]. Only one springBootApplication can be configured per Liberty server."

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/config/ServerConfigDocument.java
```diff
@@ -529,6 +528,20 @@ private void parseApplication(Document doc, XPathExpression expression) throws X
         }
     }
 
+    private void checkForSpringBootApplicationNode(Document doc, XPathExpression expression, String nodeValue) throws PluginExecutionException {
+        if(expression.equals(XPATH_SERVER_SPRINGBOOT_APPLICATION)){
+            // checking whether any springBootAppNodeLocation already configured from other server configuration files
+            if(springBootAppNodeLocation.isPresent() && springBootAppNodeDocumentURI.isPresent()){
+                throw new PluginExecutionException(String.format("Found multiple springBootApplication elements are specified in multiple server configuration files [%s, %s]. Only one springBootApplication can be configured per Liberty server.", springBootAppNodeDocumentURI.get(), doc.getDocumentURI()));
```
cherylking, line 535: "Found multiple springBootApplication elements specified in the server configuration in files [%s, %s]. Only one springBootApplication can be configured per Liberty server."

Issue #472, opened 2025-03-27
Versionless is supported since 24.0.0.8 not 24.0.0.10


krismarc:
[ERROR] Failed to execute goal io.openliberty.tools:liberty-maven-plugin:3.11.3:run (default-cli) on project spring-demo: Error installing features for server defaultServer: Detected versionless feature(s) for installation. The minimum required Liberty version for versionless feature support is 24.0.0.10 -> [Help 1]

https://github.com/OpenLiberty/ci.common/blob/dc02c551527ef1b68f523c031ddeb576ad1a9c74/src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java#L116

Accordingly to my tests and this blog info it's supproted since 24.0.0.8.
https://openliberty.io/blog/2024/08/13/24.0.0.8.html

cherylking:
@krismarc Although versionless features were introduced in 24.0.0.8, full support was not available (as in there were several issues identified that needed to be fixed in later releases) until 24.0.0.10. Therefore, we do not support the use of versionless features through the Liberty Maven and Gradle plugin unless you are using Liberty 24.0.0.10 or later.

PR #473, opened 2025-03-27
MIN_VERSIONLESS_FEATURE_VERSION too high

krismarc:
Versionless is supported since 24.0.0.8
https://github.com/OpenLiberty/ci.common/issues/472


cherylking:
@krismarc Please see my answer in the issue you opened. We need to keep the error condition when using a runtime older than 24.0.0.10.

PR #479, opened 2025-06-16
Handle generateToSrc option in dev mode

turkeylurkey:
Part of https://github.com/OpenLiberty/ci.maven/issues/1897

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -439,7 +439,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
     private Set<String> compileArtifactPaths;
     private Set<String> testArtifactPaths;
     protected File generatedFeaturesFile;
-    protected File generatedFeaturesFileParent;
+    protected File generatedFeaturesContextDir;
     protected File generatedFeaturesTmpDir;
     private File modifiedSrcBuildFile;
 
@@ -572,12 +572,12 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
     }
 
     private void initGenerationContext() {
-        this.generatedFeaturesFileParent = generateToSrc ? configDirectory : generatedFeaturesTmpDir;
-        this.generatedFeaturesFile = new File(generatedFeaturesFileParent, BinaryScannerUtil.GENERATED_FEATURES_FILE_PATH);
+        this.generatedFeaturesContextDir = generateToSrc ? configDirectory : generatedFeaturesTmpDir;
+        this.generatedFeaturesFile = new File(generatedFeaturesContextDir, BinaryScannerUtil.GENERATED_FEATURES_FILE_PATH);
     }
 
     public void copyGeneratedFeaturesFile(File destinationDir) throws IOException {
-        copyFile(generatedFeaturesFile, generatedFeaturesFileParent, destinationDir, null);
+        copyFile(generatedFeaturesFile, generatedFeaturesContextDir, destinationDir, null);
     }
     /**
      * Run unit and/or integration tests
@@ -3026,7 +3026,7 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                 registerSingleFile(containerfileUsed, executor);
             }
 
-            // Always register this hidden temp directory because generateToSrc can be turned on in dev mode
+            // Always register the generated file in the temp dir. because generateToSrc can be toggled on and off in dev mode
             File hiddenTempGenerateFeaturesFile = new File(generatedFeaturesTmpDir, BinaryScannerUtil.GENERATED_FEATURES_FILE_PATH);
             hiddenTempGenerateFeaturesFile.getParentFile().mkdirs(); // must only mkdir on the directories
             registerSingleFile(hiddenTempGenerateFeaturesFile, executor);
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -33,8 +33,10 @@ public abstract class BinaryScannerUtil {
     public static final String BINARY_SCANNER_MAVEN_VERSION = "[24.0.0.2]";
 
     public static final String GENERATED_FEATURES_FILE_NAME = "generated-features.xml";
-    public static final String GENERATED_FEATURES_FILE_PATH = "configDropins/overrides/" + GENERATED_FEATURES_FILE_NAME;
```
TrevCraw: Update copyright header in this file.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -570,8 +572,12 @@ public DevUtil(File buildDirectory, File serverDirectory, File sourceDirectory,
     }
 
     private void initGenerationContext() {
-        this.genContextDir = generateToSrc ? configDirectory : serverDirectory;
-        this.generatedFeaturesFile = new File(genContextDir, BinaryScannerUtil.GENERATED_FEATURES_FILE_PATH);
+        this.generatedFeaturesFileParent = generateToSrc ? configDirectory : generatedFeaturesTmpDir;
+        this.generatedFeaturesFile = new File(generatedFeaturesFileParent, BinaryScannerUtil.GENERATED_FEATURES_FILE_PATH);
```
TrevCraw: It probably doesn't have a negative effect... but I'm just curious if there is a particular reason that the generated file for the temp dir has to be at "configDropins/overrides" within the temp dir?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -3007,6 +3026,11 @@ public void watchFiles(File outputDirectory, File testOutputDirectory, final Thr
                 registerSingleFile(containerfileUsed, executor);
             }
 
+            // Always register this hidden temp directory because generateToSrc can be turned on in dev mode
```
TrevCraw: I don't quite follow the meaning of this comment. Also, the code below just registers the individual file and not the directory, right?

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -570,8 +572,12 @@ public DevUtil(File buildDirectory, File serverDirectory, File sourceDirectory,
...
 
     private void initGenerationContext() {
-        this.genContextDir = generateToSrc ? configDirectory : serverDirectory;
-        this.generatedFeaturesFile = new File(genContextDir, BinaryScannerUtil.GENERATED_FEATURES_FILE_PATH);
+        this.generatedFeaturesFileParent = generateToSrc ? configDirectory : generatedFeaturesTmpDir;
+        this.generatedFeaturesFile = new File(generatedFeaturesFileParent, BinaryScannerUtil.GENERATED_FEATURES_FILE_PATH);
+    }
+
+    public void copyTempFeatureFileToServer(File to) throws IOException {
+        copyFile(generatedFeaturesFile, generatedFeaturesFileParent, to, null);
```
TrevCraw: Do we need this method?

If so, we should probably rename to "copyFeatureFileToServer", since it will not always be the temp generated file, it could be generated in source.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -5862,6 +5890,7 @@ private boolean serverFeaturesModified() {
         if (generateFeatures) {
             // generateFeatures scenario: check if a generated feature has been manually added to other config files
             // Here we pass generated-features.xml instead of server.xml to calculate the generated ones
+            // The second parameter will not be used for generated-features.xml.
```
TrevCraw: Could you elaborate on this? Just for my understanding.

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: 👍

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: No particular reason except to be careful. Sometimes we pass a path around instead of a base file name. I'm not sure if this would have an effect but I wanted to keep the two ways of handling the file similar.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: The second parameter (serverDirectory) of getServerXmlFeatures() is only used if the third parameter (generated-features.xml) contains an `<include>` element to bring in more files. Technically the 2nd parameter is used in a logging stmt but that doesn't matter functionally. The point is that it doesn't really matter what value the second parameter has, we don't need to sweat. Unless you're debugging and the log stmt comes out wrong. But our generated file is not going to have an `<include>`.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
turkeylurkey: I meant "off"

PR #487, opened 2025-10-03
adding port info to help message

arunvenmany-ibm:
Fixes https://github.com/OpenLiberty/ci.maven/issues/1447

image

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2485,78 +2485,82 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
             hotTests = true;
         }
         if (startup) {
-            if (container) {
-                boolean nonDefaultHttpPortUsed = !skipDefaultPorts && !String.valueOf(LIBERTY_DEFAULT_HTTP_PORT).equals(httpPort);
-                boolean nonDefaultHttpsPortUsed = !skipDefaultPorts && !String.valueOf(LIBERTY_DEFAULT_HTTPS_PORT).equals(httpsPort);
-                boolean nonDefaultDebugPortUsed = alternativeDebugPort != -1; // this is set when a random ephemeral port is selected
-                if (containerHttpPort != null || containerHttpsPort != null || libertyDebug) {
-                    info(formatAttentionMessage(""));
-                    info(formatAttentionTitle("Liberty container port information:"));
-                }
-                if ((containerHttpPort != null && httpPort != null && nonDefaultHttpPortUsed)
-                        || (containerHttpsPort != null && httpsPort != null && nonDefaultHttpsPortUsed)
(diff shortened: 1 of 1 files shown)


cherylking:
Two things...can I see a before and after image..not just after? Also, is this only for when they ask for help with `h`? That is what the original issue was asking for.

cherylking:
I think I prefer image 1. I wonder if we should push all the port info to the right though? Seems odd that `Liberty server port information` is aligned to the left with nothing else.

arunvenmany-ibm:
> I think I prefer image 1. I wonder if we should push all the port info to the right though? Seems odd that `Liberty server port information` is aligned to the left with nothing else.

Thats actually the title.. its aligned similarly in case of startup message as well

image

arunvenmany-ibm:
> Still looks weird...maybe we should consider the other proposal in that issue which was to provide an option to display the port info....could use `p` for it. Help usually is showing info on options...not details about the server. If we think there might be other details about the server to display, we could use `i` and say display information about the Liberty server. If we use `p`, then display port information for the Liberty server.

ok.. using 'p'

when liberty dev mode is started, option p is not shown as liberty port information is already being displayed -> do you think we should display option 'p' here as well ?
image


when user selects h for help, a message is displayed with the new option 'p'
image

when user press 'p' , port information is shown

image

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2775,6 +2778,14 @@ private void readInput() {
                         } else {
                             runTestThread(false, executor, -1, true, buildFile);
                         }
+                    }else if (p.isPressed(line) ) {
```
cherylking, line 2781: I think this should be before the check for `t` because it also checks for `enter`. Also fix the spacing for `else if`.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2775,6 +2778,14 @@ private void readInput() {
...
+                        try {
+                            info(formatAttentionBarrier());
+                            printPortInfo(true);
+                            info(formatAttentionBarrier());
+                        } catch (PluginExecutionException ignored) {
+                            // ignoring for help
```
cherylking, line 2787: The other sections above do not have try/catch...why does this one?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
arunvenmany-ibm: Done

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
arunvenmany-ibm: because printPortInfo throws PluginExecutionException.. Instead of ignore changed to logging error

cherylking (approved):

PR #501, opened 2026-02-25
adding JAVA_HOME env variable for product info based on java.home

venmanyarun:
Fixing https://github.com/OpenLiberty/ci.gradle/issues/1060

Similar code is already present in ci.ant for server task and we are adding for ProductInfo

https://github.com/OpenLiberty/ci.ant/blob/26245d3e002293b4babc1efde6f3ab7089a51521/src/main/java/io/openliberty/tools/ant/ServerTask.java#L113

image

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
@@ -1151,7 +1151,14 @@ public abstract class InstallFeatureUtil extends ServerFeatureUtil {
             } else {
                 productInfoFile = installDirectory + "/bin/productInfo";
             }
+
             ProcessBuilder pb = new ProcessBuilder(productInfoFile, action);
+            Properties sysp = System.getProperties();
+            String javaHome = sysp.getProperty("java.home");
+            if (javaHome != null) {
+                pb.environment().put("JAVA_HOME", javaHome);
+            }
+            pb.redirectErrorStream(true);
             pr = pb.start();
 
             in = new BufferedReader(new InputStreamReader(pr.getInputStream()));


src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
```diff
@@ -1151,7 +1151,14 @@ public static String productInfo(File installDirectory, String action) throws Pl
             } else {
                 productInfoFile = installDirectory + "/bin/productInfo";
             }
+
             ProcessBuilder pb = new ProcessBuilder(productInfoFile, action);
+            Properties sysp = System.getProperties();
```
cherylking, line 1156: I'm a little concerned this could ignore whatever has been set up with Java toolchains. Wondering if we should only do this when there doesn't seem to be any JAVA_HOME set already?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/InstallFeatureUtil.java
venmanyarun: Currently we have not incorparated toolchains for this action

I have created an issue for this
https://github.com/OpenLiberty/ci.common/issues/502

cherylking: And what about for the server task that you referenced in ci.ant above? Do we need code changes there? I don't think we should add this logic here without accounting for Java toolchains.

venmanyarun: ci.ant task for InstallLibertyTask  is not referring to productInfo/ server
its direct java code . We do not need code changes there

cherylking (approved):
I'm approving this PR, but I want the other issue resolved before we do another release.

PR #505, opened 2026-04-06
Duplicate recompilation fixed for java files

sajeerzeji:
Fixes an issue mentioned in the comments of [Maven:#752](https://github.com/OpenLiberty/ci.maven/issues/752), where the issue was about the processFileChanges() was being called multiple times for the same .properties file when modifying in dev mode, but that is working correctly now, but I found an issue that is processFileChanges() is getting called twice for .class files when we change Java sources in dev mode.

When a .java file changes, the plugin queues it for compilation in the recompileJavaSources list. After a timeout, the plugin compiles the .java file into a .class file. Liberty then detects this .class file change and hot reloads the application.

The problem was that the .java file remained in the recompileJavaSources queue even after being compiled. This caused the plugin to compile the same .java file a second time, creating duplicate work and triggering unnecessary file change events.

The fix detects when Liberty has successfully hot reloaded by checking if the Application Update message count increased. If it has reloaded, we clear the recompileJavaSources list to prevent the plugin from compiling the same .java files again. This eliminates the duplicate compilation thus the processFileChanges() will be called only once.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -4522,6 +4522,17 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                 debug("Java source class file modified: " + fileChanged.getName()
                         + ". Adding to list for processing.");
                 modifiedClasses.add(fileChanged);
+
+                if (!recompileJavaSources.isEmpty()) {
+                    int currentMessages = countApplicationUpdatedMessages();
+                    if (currentMessages > numApplicationUpdatedMessages) {
+                        debug("Liberty hot reload detected (CWWKZ0003I), clearing recompileJavaSources list to prevent duplicate recompilation");
+                        debug("Files that will not be recompiled: " + recompileJavaSources);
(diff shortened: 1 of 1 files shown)


venmanyarun (changes requested):
Please create PR's for ci.maven and ci.gradle with updated tests to make sure integration-tests are working fine

sajeerzeji:
> Please create PR's for ci.maven and ci.gradle with updated tests to make sure integration-tests are working fine

Added PRs:

https://github.com/OpenLiberty/ci.gradle/pull/1075
https://github.com/OpenLiberty/ci.maven/pull/1997

venmanyarun (approved):

venmanyarun:
@sajeerzeji 
Add a screen recording of the issue fix here, ass the case is IDE specific

sajeerzeji:
> @sajeerzeji Add a screen recording of the issue fix here, ass the case is IDE specific

Before:
https://github.com/user-attachments/assets/6b75f770-176c-4f88-9536-589a051b35fc

After:

https://github.com/user-attachments/assets/97855ca9-1e34-47b3-adcb-2f73cc1459a4

PR #506, opened 2026-04-07
Support a new API in the binary scanner

turkeylurkey:
Add strings that define the Maven coordinates needed to download Liberty feature lists for available releases.
Accept feature list files to be passed to the binary scanner using a new API. 
Rename a variable which represents a jar file from the originally confusing name.

Related to https://github.com/OpenLiberty/ci.maven/issues/1981

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
@@ -36,31 +36,25 @@ public abstract class BinaryScannerUtil {
     public static final String OLBASE_FEATURELIST_GROUP_ID = "io.openliberty.features";
     public static final String OLBASE_FEATURELIST_ARTIFACT_ID = "open_liberty_featurelist";
     public static final String OLBASE_FEATURELIST_TYPE = "xml";
-    public static final String OLBASE_FEATURELIST_VERSION = "[25.0.0.7]";
+    // the version number is generated at the point of use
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -30,7 +30,31 @@ public abstract class BinaryScannerUtil {
     public static final String BINARY_SCANNER_MAVEN_GROUP_ID = "com.ibm.websphere.appmod.tools";
     public static final String BINARY_SCANNER_MAVEN_ARTIFACT_ID = "binary-app-scanner";
     public static final String BINARY_SCANNER_MAVEN_TYPE = "jar";
-    public static final String BINARY_SCANNER_MAVEN_VERSION = "[25.0.0.2]";
+    public static final String BINARY_SCANNER_MAVEN_VERSION = "[25.0.0.2.1]";
+
+    // The coordinates to use for Open Liberty versions 25.0.0.7 and up
+    public static final String OLBASE_FEATURELIST_GROUP_ID = "io.openliberty.features";
+    public static final String OLBASE_FEATURELIST_ARTIFACT_ID = "open_liberty_featurelist";
+    public static final String OLBASE_FEATURELIST_TYPE = "xml";
```
TrevCraw: I would recommend removing "base" from the OL feature list constants since there is no "base" or "core" for OL.

turkeylurkey: Agreed, kind of a blind spot.

TrevCraw (approved):

PR #508, opened 2026-04-20
Updated user facing output

TrevCraw:
- Removed warnings in the help menu regarding writing to source dir.
- Changed to "write to source" warning message to an info message.


turkeylurkey (approved):
Looks good.

PR #509, opened 2026-04-21
Use the binary scanner API with a Map

turkeylurkey:
Part of https://github.com/OpenLiberty/ci.maven/issues/2001

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
@@ -31,7 +31,7 @@ public abstract class BinaryScannerUtil {
     public static final String BINARY_SCANNER_MAVEN_GROUP_ID = "com.ibm.websphere.appmod.tools";
     public static final String BINARY_SCANNER_MAVEN_ARTIFACT_ID = "binary-app-scanner";
     public static final String BINARY_SCANNER_MAVEN_TYPE = "jar";
-    public static final String BINARY_SCANNER_MAVEN_VERSION = "[26.0.0.2-20260420]";
+    public static final String BINARY_SCANNER_MAVEN_VERSION = "[25.0.0.2.2]";
 
     // The coordinates to use for Open Liberty versions 25.0.0.7 and up
     public static final String OL_FEATURELIST_GROUP_ID = "io.openliberty.features";
@@ -154,7 +154,7 @@ public abstract class BinaryScannerUtil {
             String logLocation, String targetJavaEE, String targetMicroProfile, Map featureListFileMap, boolean optimize)
             throws PluginExecutionException, NoRecommendationException, RecommendationSetException, FeatureModifiedException,
             FeatureUnavailableException, IllegalTargetException, IllegalTargetComboException {
-        Set<String> featureList = null;
+        Set<String> generatedFeatureList = null;
         if (binaryScannerJar != null && binaryScannerJar.exists()) {
             // if we are already generating features for all class files (optimize=true) and
             // we are not passing any user specified features (currentFeatureSet is empty)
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -170,13 +176,21 @@ public Set<String> runBinaryScanner(Set<String> currentFeatureSet, List<String>
                         "  targetJavaEE: " + targetJavaEE + "\n" +
                         "  targetMicroP: " + targetMicroProfile + "\n" +
                         "  currentFeatures: " + currentFeatureSet + "\n" +
-                        "  baseFeatureListFile: " + ((baseFeatureListFile == null) ? "null" : baseFeatureListFile.getAbsolutePath()) + "\n" +
-                        "  coreFeatureListFile: " + ((coreFeatureListFile == null) ? "null" : coreFeatureListFile.getAbsolutePath()) + "\n" +
+                        "  featureListFileMap: " + featureListFileMap + "\n" +
+                        "  logLocation: " + logLocation + "\n" +
+                        "  logLevel: " + logLevel + "\n" +
+                        "  locale: " + java.util.Locale.getDefault());
+                warn ("Calling " + binaryScannerJar.getName() + " with the following inputs...\n" +
```
TrevCraw: I think you forgot to remove the extra logging section you were using for debugging.

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -145,7 +151,7 @@ public BinaryScannerUtil(File binaryScanner) {
      *                                       scanner when used in combination with each other. E.g. EE 7 and MP 2.1
      */
     public Set<String> runBinaryScanner(Set<String> currentFeatureSet, List<String> classFiles, Set<String> allClassesDirectories,
-            String logLocation, String targetJavaEE, String targetMicroProfile, File baseFeatureListFile, File coreFeatureListFile, boolean optimize)
+            String logLocation, String targetJavaEE, String targetMicroProfile, Map featureListFileMap, boolean optimize)
             throws PluginExecutionException, NoRecommendationException, RecommendationSetException, FeatureModifiedException,
             FeatureUnavailableException, IllegalTargetException, IllegalTargetComboException {
         Set<String> featureList = null;
```
TrevCraw: It is now occurring to me that with the latest changes using the Liberty feature lists, "featureList" is an overloaded term. I'm thinking it might make things clearer if we rename this variable to make it more explicit as `generatedFeatureList`.

TrevCraw (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: I noticed that too but I didn't check in a fix because the version number still needs to be updated.

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: No, these are the lists of features that we download from Maven central so they are the complete list of all the Liberty features in base or core.

TrevCraw: I think your comment may be proving my point :)

I am talking about the `Set<String> featureList` variable declared on this line: https://github.com/turkeylurkey/ci.common/blob/ab17bc34b728db20e99e415f10fde1b990088b11/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java#L157
...and assigned to the list returned from the binary scanner method call on this line: https://github.com/turkeylurkey/ci.common/blob/ab17bc34b728db20e99e415f10fde1b990088b11/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java#L192

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
TrevCraw: We may need to move forward before a binary scanner is available in Maven Central.

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
turkeylurkey: Oh! Ok.

TrevCraw (approved):

PR #511, opened 2026-04-21
Merge feature branch for auto-generation of Liberty features GA update

TrevCraw:


src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -23,17 +23,50 @@
...
+import java.util.Map;
 import java.util.Set;
 
 public abstract class BinaryScannerUtil {
 
     public static final String BINARY_SCANNER_MAVEN_GROUP_ID = "com.ibm.websphere.appmod.tools";
     public static final String BINARY_SCANNER_MAVEN_ARTIFACT_ID = "binary-app-scanner";
     public static final String BINARY_SCANNER_MAVEN_TYPE = "jar";
-    public static final String BINARY_SCANNER_MAVEN_VERSION = "[25.0.0.2,)";
+    public static final String BINARY_SCANNER_MAVEN_VERSION = "[25.0.0.2.2]";
```
venmanyarun: are we going to use a fixed version always?

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -252,9 +286,9 @@ public Set<String> runBinaryScanner(Set<String> currentFeatureSet, List<String>
      * @return - a set of features that will allow the application to run in a Liberty server
      * @throws PluginExecutionException - any exception that prevents the scanner from running
      */
-    public Set<String> reRunBinaryScanner(Set<String> allClassesDirectories, String logLocation, String targetJavaEE, String targetMicroProfile)
-            throws PluginExecutionException {
-        Set<String> featureList = null;
+    public Set<String> reRunBinaryScanner(Set<String> allClassesDirectories, String logLocation, String targetJavaEE, String targetMicroProfile,
```
venmanyarun: I can see that runBinaryScanner and rerunBinaryScanner handle the same binary scanner exceptions completely differently
is this intentional ?

example
PROVIDED_FEATURE_EXCEPTION 
runBinaryScanner approach:

1. Throws NoRecommendationException if no alternatives found
2. Throws RecommendationSetException if alternatives exist

rerunBinaryScanner approach:

Logs debug message: "RuntimeException from re-run of binary scanner"
Returns null

venmanyarun (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
TrevCraw: per release, yes it will be a fixed version

src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
TrevCraw: Yes, this is intentional. The contexts to call `runBinaryScanner` and `rerunBinaryScanner` are different. We only call `rerunBinaryScanner` when there was an error from the binary scanner with the initial `runBinaryScanner` call. The second call to binary scanner in `rerunBinaryScanner` is intended to get a pure binary scanner recommended feature list for the application (i.e., analyze the whole application and ignore user-specified features). Because the contexts are different, the exceptions from the binary scanner are handled differently.

venmanyarun (approved):

venmanyarun (approved):

PR #514, opened 2026-05-04
Reorganize the Maven ccordinates used to download the Websphere feature lists

turkeylurkey:
Part of https://github.com/OpenLiberty/ci.maven/issues/2010

Put nonsense value for binary scanner version.

Add the version numbers for the different Maven coordinates previously recorded in ci.maven

Organize the Maven coordinates into three groups, called cadres so we wont confuse them with "groupid".

We could encapsulate the complexity of the Maven coordinates in the code that fetches the file but I decided to put the complexity in the data seen here.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
@@ -42,23 +42,23 @@ public abstract class BinaryScannerUtil {
     public static final String OL_FEATURELIST_KEY = "openLiberty";
 
     // The coordinates to use for WebSphere Liberty versions describe three different
-    // cadres. Version 25.0.0.7 and earlier are the first cadre. Version 25.0.0.8 through
-    // to 25.0.0.12 are the second cadre. The last cadre is versions 26.0.0.1 and onward.
-    // Maven coordinates for cadre 1
-    public static final String WS_FEATURE_LIST_VERSION_CADRE1 = "25.0.0.7";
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/BinaryScannerUtil.java
```diff
@@ -41,27 +41,30 @@ public abstract class BinaryScannerUtil {
...
-    // These coordinates are used in different combinations to access WebSphere Liberty feature lists 25.0.0.7 and up
-    public static final String WS1_FEATURELIST_GROUP_ID = "io.openliberty.features";
-    public static final String WS2_FEATURELIST_GROUP_ID = "com.ibm.websphere.appserver.features";
-    public static final String WSBASE_FEATURELIST_ARTIFACT_ID = "websphere_liberty_base__featurelist";
-    public static final String WSCORE_FEATURELIST_ARTIFACT_ID = "websphere_liberty_core__featurelist";
+    // The coordinates to use for WebSphere Liberty versions describe three different
+    // cadres. Version 25.0.0.7 and earlier are the first cadre. Version 25.0.0.8 through
+    // to 25.0.0.12 are the second cadre. The last cadre is versions 26.0.0.1 and onward.
+    // Maven coordinates for cadre 1
+    public static final String WS_FEATURE_LIST_VERSION_CADRE1 = "25.0.0.7";
```
TrevCraw: Would "set" or "batch" be better options than "cadre"?

turkeylurkey: I think "batch" is best since people might confuse it with a Java Set.

TrevCraw (approved):

PR #517, opened 2026-05-15
changes to start hotkey reader earlier before server is fully started

venmanyarun:
Fixing https://github.com/OpenLiberty/ci.maven/issues/1638

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2826,10 +2826,14 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                             earlyQuitRequested.set(true);
                         }
                         runShutdownHook(executor);
+                    } else if (h.isPressed(line)) {
+                        info(formatAttentionBarrier());
+                        printHelpMessages();
+                        info(formatAttentionBarrier());
                     } else if (serverStarting.get()) {
-                        // During server startup, only 'q' is allowed. Ignore all other keys.
+                        // During server startup, only 'q' and 'h' are allowed. Ignore all other keys.
                         if (!line.trim().isEmpty()) {
-                            info("Server is starting. Only 'q' (quit) is available. Please wait for server to start.");
+                            info("The requested command is not available during server startup.");
                         }
                     } else if (r.isPressed(line)) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2773,6 +2820,11 @@ private void readInput() {
                     String line = scanner.nextLine();
                     if (q.isPressed(line)) {
                         debug("Detected exit command");
+                        // If server is still starting, set the early quit flag
```
cherylking, line 2823: What happens if one of the other hot keys is requested early? Shouldn't we prevent that?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
venmanyarun: updated code to fix this. added tests in ci.common and ci.maven as well

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2826,6 +2826,11 @@ private void readInput() {
                             earlyQuitRequested.set(true);
                         }
                         runShutdownHook(executor);
+                    } else if (serverStarting.get()) {
+                        // During server startup, only 'q' is allowed. Ignore all other keys.
+                        if (!line.trim().isEmpty()) {
+                            info("Server is starting. Only 'q' (quit) is available. Please wait for server to start.");
```
cherylking: We don't say "please" in end user messages.

"The requested command is not available during server startup." or
"The requested command is not available while the server is starting."

I think that would be sufficient.

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
cherylking: btw...we could probably let them use `h` as well?

venmanyarun: Done.. message changed and code changes made to allow "h"

cherylking (approved):

PR #519, opened 2026-05-29
Added support for passing annotation processor paths to the Java compiler

sajeerzeji:
Fixes https://github.com/OpenLiberty/ci.maven/issues/2021

Added support for passing annotation processor paths to the Java compiler via the -processorpath option. This enables annotation processors (Eg: Lombok) to run during compilation in dev mode.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/JavaCompilerOptions.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/JavaCompilerOptions.java
@@ -47,6 +47,11 @@ public class JavaCompilerOptions {
     private String encoding;
 
     /**
+     * Annotation processor path
+     */
+    private String annotationProcessorPath;
+
+    /**
      * Get list of options that can be passed to the compiler.
      * Options with values are represented as multiple strings in the list
      *  e.g. "-source" and "1.8"
@@ -62,6 +67,7 @@ public class JavaCompilerOptions {
         addStringOption(options, "-target", target);
         addStringOption(options, "--release", release);
         addStringOption(options, "-encoding", encoding);
+        addStringOption(options, "-processorpath", annotationProcessorPath);
         return options;
     }
 
@@ -110,6 +116,14 @@ public class JavaCompilerOptions {
 
     public void setEncoding(String encoding) {
         this.encoding = encoding;
+    }
+
+    public String getAnnotationProcessorPath() {
+        return annotationProcessorPath;
+    }
+
+    public void setAnnotationProcessorPath(String annotationProcessorPath) {
+        this.annotationProcessorPath = annotationProcessorPath;
     }
 
 }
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/JavaCompilerOptions.java
```diff
@@ -62,6 +67,7 @@ public List<String> getOptions() {
         addStringOption(options, "-target", target);
         addStringOption(options, "--release", release);
         addStringOption(options, "-encoding", encoding);
+        addStringOption(options, "-processorpath", annotationProcessorPath);
```
cherylking, line 75: Should we also support the `annotationProcessors` parameter? When I look at ExecuteMojoUtil COMPILE_PARAMS, I see a very long list of possible parameters. Are there more in that list that should be passed along in the JavaCompilerOptions?

cherylking (changes requested):

src/main/java/io/openliberty/tools/common/plugins/util/JavaCompilerOptions.java
sajeerzeji: Added annotation processor path and processors support to JavaCompilerOptions for dev mode hot reload compilation

cherylking: What about all the other parameters? I am not sure why we have them listed in COMPILE_PARAMS if we are not actually handling them.

sajeerzeji: I checked this concern, the COMPILE_PARAMS is used to filter the maven-compiler-plugin configuration before executing the plugin. JavaCompilerOptions is used for hot reload and is populated by getMavenCompilerOptions() and passed to recompileJava().

Most COMPILE_PARAMS are not apply to JavaCompilerOptions, plugin specific parameters(compilerArgs, compilerArgument, etc) control the plugin not the compiler, common parameters(source, target, encoding) are already extracted, and build lifecycle parameters are not relevant for hot reload I think. Annotation processor parameters were the missing ones here, they need artifact resolution and are critical for hot reload, especially for libraries like Lombok.

cherylking: Ok, thank you for researching Sajeer. I just wanted to make sure while we were here fixing the code that we covered everything we should.

src/main/java/io/openliberty/tools/common/plugins/util/JavaCompilerOptions.java
```diff
@@ -46,6 +46,16 @@ public class JavaCompilerOptions {
      */
     private String encoding;
 
+    /**
```
cherylking, line 49: Update copyright year to include 2026. The proper way is you leave the 2020 and just add `, 2026`.

cherylking (changes requested):

cherylking (approved):

src/main/java/io/openliberty/tools/common/plugins/util/JavaCompilerOptions.java
sajeerzeji: Added year 2026 to the copyright header

PR #520, opened 2026-06-01
Refactor some feature gen variables

turkeylurkey:
Embrace the feature gen.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/FeatureGenUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/FeatureGenUtil.java
@@ -26,19 +26,19 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-public abstract class BinaryScannerUtil {
+public abstract class FeatureGenUtil {
 
-    public static final String BINARY_SCANNER_MAVEN_GROUP_ID = "com.ibm.websphere.appmod.tools";
-    public static final String BINARY_SCANNER_MAVEN_ARTIFACT_ID = "feature-gen";
-    public static final String BINARY_SCANNER_MAVEN_TYPE = "jar";
-    public static final String BINARY_SCANNER_MAVEN_VERSION = "[0.0.1-SNAPSHOT,)"; // use the specified version or later
+    public static final String FEATURE_GEN_MAVEN_GROUP_ID = "com.ibm.websphere.appmod.tools";
+    public static final String FEATURE_GEN_MAVEN_ARTIFACT_ID = "feature-gen";
+    public static final String FEATURE_GEN_MAVEN_TYPE = "jar";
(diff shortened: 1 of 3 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/FeatureGenUtil.java
```diff
@@ -26,19 +26,19 @@
 import java.util.Map;
 import java.util.Set;
 
-public abstract class BinaryScannerUtil {
+public abstract class FeatureGenUtil {
```
TrevCraw: I just want to think through this name change a bit. We are looking to consolidate the common code for [GenerateFeaturesMojo.java](https://github.com/OpenLiberty/ci.maven/blob/main/liberty-maven-plugin/src/main/java/io/openliberty/tools/maven/server/GenerateFeaturesMojo.java) and [GenerateFeaturesTask.groovy](https://github.com/OpenLiberty/ci.gradle/blob/main/src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy). I assume the likely name for the class/file that contains the common code would be `GenerateFeaturesUtil`. If that is the case, we would have `GenerateFeaturesUtil` and `FeatureGenUtil`. I think it would be hard to differentiate the meaning of those file names. Should we use something like `FeatureGenDriverUtil`, `FeatureGenJarUtil`, or `FeatureGeneratorUtil` to show a clear distinction?

turkeylurkey: since we have the method `runFeatureGenerator()` I think `FeatureGeneratorUtil` is the best choice.

src/main/java/io/openliberty/tools/common/plugins/util/FeatureGeneratorUtil.java
```diff
@@ -193,111 +193,111 @@ public Set<String> runBinaryScanner(Set<String> currentFeatureSet, List<String>
...
                 // 4. FeatureNotAvailableAtRequestedLevelException means the features passed or binary files
                 // scanned require features that do not exist at the requested EE or MP levels.
                 // 5. IllegalTargetException means that the Java or Jakarta EE version or the MicroProfile version
-                // we read from the build file is out of range for the binary scanner. For EE we only use the first
+                // we read from the build file is out of range for the feature generator. For EE we only use the first
                 // digit: ee6 to ee9. For MP we use the first two digits mp1.2 to mp5.0.
                 // 6. IllegalTargetCombinationException means that the EE level and the MP level are not compatible.
-                Throwable scannerException = ite.getCause();
-                if (scannerException.getClass().getName().equals(PROVIDED_FEATURE_EXCEPTION)) {
+                Throwable featureGenException = ite.getCause();
```
TrevCraw: suggestion to change to `generatorException`

turkeylurkey: 👍

TrevCraw (approved):

PR #523, opened 2026-06-16
Detect files to ignore in generate features temp dir

turkeylurkey:
The original comment suggests we only expect to see changes to `generated-features.xml` in the temp dir. Add code to enforce this so we do not process other files that are copied there by necessity during ~install~ generate features. 

Fixes: https://github.com/OpenLiberty/ci.maven/issues/2046

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -4514,8 +4514,11 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
         } else if ((directory.startsWith(configPath)
                 || directory.startsWith(gfTmpDirPath))
                 && !isGeneratedConfigFile(fileChanged, configDirectory, serverDirectory)) {
-            // configuration file or generate-features.xml in temp directory
-            processConfigFileChange(fileChanged, changeType, executor, numApplicationUpdatedMessages, false);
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -4514,8 +4514,11 @@ private void processFileChanges(
         } else if ((directory.startsWith(configPath)
                 || directory.startsWith(gfTmpDirPath))
                 && !isGeneratedConfigFile(fileChanged, configDirectory, serverDirectory)) {
```
TrevCraw, line 4517: Wouldn't it be cleaner to modify the original statement? Just add an AND statement for the file name to `directory.startsWith(gfTmpDirPath)`. It would look something like below:

```suggestion
} else if ((directory.startsWith(configPath)
|| (directory.startsWith(gfTmpDirPath) && fileChanged.getName().equals(FeatureGeneratorUtil.GENERATED_FEATURES_FILE_NAME)))
&& !isGeneratedConfigFile(fileChanged, configDirectory, serverDirectory)) {
```

turkeylurkey: I was thinking the if/else-if format was breaking the files into coarse grained groups and then we could further sort through the files inside each group. But the `isGeneratedConfigFile()` kind of breaks that concept. What you propose makes sense taking that method into account.

TrevCraw (approved):

PR #526, opened 2026-07-31
Fixed the race condition: print dev mode ready message after WatchService registration

sajeerzeji:
Fixes https://github.com/OpenLiberty/ci.maven/issues/2070

PR Ci.Maven: https://github.com/OpenLiberty/ci.maven/pull/2073
PR Ci.Gradle: https://github.com/OpenLiberty/ci.gradle/pull/1095

In DevUtil.watchFiles(), the "Liberty is running in dev mode." startup banner is now printed after all source, test, config, and resource directories have been registered with the WatchService, but before initWatchLoop() enters the event poll. Previously the banner was emitted by runHotkeyReaderThread() in the callers (ci.maven's DevMojo, ci.gradle's DevTask) before watchFiles() was even called, meaning any file change made immediately after that message by a test or a user could race ahead of the watcher and be silently dropped with no recompile triggered. Moving the print into watchFiles() and checking it in with the existing firstStartup flag makes the ready message a reliable synchronisation point: once a user or test sees it, the WatchService is fully set up and guaranteed to detect changes. The runHotkeyReaderThread() post test run re print path is unchanged.

### DIFF
--- a/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
+++ b/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
@@ -2548,8 +2548,8 @@ public abstract class DevUtil extends AbstractContainerSupportUtil {
                     // it's available
                     inputUnavailable.wait(500);
                 }
-                printDevModeMessages(inputUnavailable.get(), firstStartup);
-                firstStartup = false;
+                // firstStartup is false by the time this is called after initial startup.
+                printDevModeMessages(inputUnavailable.get(), false);
             } catch (InterruptedException e) {
                 debug("Interrupted while waiting to determine whether input can be read", e);
             } catch (PluginExecutionException pe) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
```diff
@@ -2548,8 +2548,8 @@ public void runHotkeyReaderThread(ThreadPoolExecutor executor) {
                     // it's available
                     inputUnavailable.wait(500);
                 }
-                printDevModeMessages(inputUnavailable.get(), firstStartup);
-                firstStartup = false;
```
turkeylurkey, line 2552: Could you explain why we don't need to clear this boolean anymore? The comment says "If the thread is already running..." so it makes me wonder if firstStartup is always false by the time this is called.

turkeylurkey (commented):
.

src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java
sajeerzeji: The firstStartup is always false by the time runHotkeyReaderThread() is reached. watchFiles() sets firstStartup = false before entering the watch loop via initWatchLoop(). The only call site for runHotkeyReaderThread() in ci.common is inside TestJob.run(), which runs after the watch loop is active, so watchFiles() has always completed by then. The old code reset it here to cover the premature call from DevMojo/DevTask before watchFiles() ran. That premature call is the race condition we are removing, and without it the reset is no longer needed.

